PR #1, opened 2018-07-07
sbi-doc PR (Updated)

atishp04:
I have tried to create the SBI spec as per my knowledge and understanding RISC-V system. There may be some glaring mistakes either in language or technical content. Apologies in advance :).

As my knowledge on memory model is very limited, I have not described the memory ordering related functions. Kindly update.

Please feel free to edit/comment on anything and everything.

### DIFF
--- a/riscv-sbi.md
+++ b/riscv-sbi.md
@@ -26,11 +26,13 @@ https://creativecommons.org/licenses/by/4.0/.
 
 ## Introduction
 
-Supervisor Binary Interface(SBI) describes the instructions that can be executed together with a set of SBI calls out to the supervisor execution environment (SEE)
+Supervisor Binary Interface (SBI) describes the instructions that can be executed
+together with a set of SBI calls out to the supervisor execution environment (SEE)
 on a given platform. Several instructions that might normally be handled by the
 supervisor operating system (OS) directly are handled via SBI calls to the SEE in
-RISC-V. This provides a cleaner interface for the supervisor OS. Currently,
-The next section will describe different types of SBI calls defined for RISC-V.
+RISC-V. This provides a cleaner interface for the supervisor OS. Currently, SBI
+is currently implemented by Berkeley Boot Loader (BBL). The next section will
+describe different types of SBI calls defined for RISC-V.
 
 ## SBI List<a name=sbi-list>
 There are mainly 5 types of SBI calls defined in the specification as per below
@@ -45,7 +47,15 @@ table.
 | Shutdown         |	sbi_shutdown |
 
 ## Description
-Each SBI call adheres to a specific calling convention that defines how data is provided as input to or read as output. Individual SBI call signature and its
+Each SBI call adheres to a specific calling convention that defines how data is
+provided as input to or read as output. The SBI calling convention matches the
+syscall calling convention (which in turn is very similar to the function calling
+convention). Every SBI call uses *ecall* instructions to make a request to the SEE
+environment. When executed in S-mode, it generates an environment-call-from-S-mode
+exception and performs no other operation. The required arguments are passed
+through registers *a0-a2* and the SBI call type is passed via register *a7*. Once
+the machine mode receives the trap, it identifies the type of SBI call from *a7*
+register and perform required operation. Individual SBI call signature and its
 purpose is described next.
 
 ### Timer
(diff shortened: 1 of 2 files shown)


atishp04:
ping ?

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+Programs the clock for next event after *stime_value* time. This function also
+clears the pending timer interrupt bit. If the supervisor needs to clear the
+timer interrupt without scheduling the next timer event, they may choose to do
+so in timer interrupt handler or any other place in supervisor.
+
+### IPI
+```C
+void sbi_send_ipi(const unsigned long *hart_mask)
+```
+Send an inter-processor interrupt to all the harts defined in hart_mask.
```
aswaterman: Suggest adding the following text:

"Interprocessor interrupts manifest at the receiving harts as Supervisor Software Interrupts.

hart_mask is a physical address that points to a bit-vector of harts.  The bit vector is represented as a sequence of unsigned longs whose length equals the number of harts in the system divided by the number of bits in an unsigned long, rounded up to the next integer.

If hart_mask is NULL, an IPI is sent to every hart in the system."

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+
+```C
+void sbi_clear_ipi(void)
+```
+Clears any pending IPIs in the hart for which this SBI is invoked.
```
aswaterman: I would say "Clears the pending IPI, if any," to make it clear only one IPI can be pending.

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Memory Model<a name=#memory-model>
+```C
+void sbi_remote_fence_i(const unsigned long *hart_mask)
+```
+Instruct remote harts to execute FENCE.I instruction.
```
aswaterman: "hart_mask is as described in sbi_send_ipi."

aswaterman:
Good start!  Do you want to address my feedback now, or in a follow-up PR?

riscv-sbi.md
atishp04: Sure.

riscv-sbi.md
atishp04: Sure. Thanks for the clarification

riscv-sbi.md
atishp04: ok

atishp04:
I will update this PR addressing your comments.

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+ &copy; 2018 Andrew Waterman <aswaterman@gmail.com>
+ &copy; 2018 Palmer Dabbelt <palmer@dabbelt.com>
+
+It is licensed under the Creative Commons Attribution 4.0 International
+License (CC-BY 4.0).  The full license text is available at
+https://creativecommons.org/licenses/by/4.0/.
+
+## Introduction
+
+Supervisor Binary Interface(SBI) describes the instructions that can be executed together with a set of SBI calls out to the supervisor execution environment (SEE)
```
palmer-dabbelt: missing space

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+on a given platform. Several instructions that might normally be handled by the
+supervisor operating system (OS) directly are handled via SBI calls to the SEE in
+RISC-V. This provides a cleaner interface for the supervisor OS. Currently,
```
palmer-dabbelt: trailing "currently,"

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+| --------------|-------------------|
+| Timer         | sbi_set_timer     |
+| IPI           | sbi_clear_ipi<br>sbi_send_ipi  |
+| Memory Model| sbi_remote_fence_i<br>sbi_remote_sfence_vma<br>sbi_remote_sfence_vma_asid
+|	Console				| sbi_console_putchar <br> sbi_console_getchar |
+| Shutdown         |	sbi_shutdown |
+
+## Description
+Each SBI call adheres to a specific calling convention that defines how data is provided as input to or read as output. Individual SBI call signature and its
+purpose is described next.
```
palmer-dabbelt: The SBI calling convention matches the syscall calling convention (which in turn is very similar to the function calling convention).

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Timer
+```C
+void sbi_set_timer(uint64_t stime_value)
+```
+Programs the clock for next event after *stime_value* time. This function also
+clears the pending timer interrupt bit. If the supervisor needs to clear the
+timer interrupt without scheduling the next timer event, they may choose to do
+so in timer interrupt handler or any other place in supervisor.
```
palmer-dabbelt: I'd just say "by modifying SIE directly".

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+```
+//TODO
+
+```C
+void sbi_remote_sfence_vma_asid(const unsigned long *hart_mask,
+                                unsigned long start,
+                                unsigned long size,
+                                unsigned long asid)
+```
+//TODO
```
palmer-dabbelt: These just look like remote_fence_i

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Console
+```C
+void sbi_console_putchar(int ch)
+```
+Write data present in *ch* to debug console (blocking)
```
palmer-dabbelt: If the console doesn't exist then the character is thrown away.

riscv-sbi.md
```diff
@@ -0,0 +1,113 @@
...
+int sbi_console_getchar(void)
+```
+Read a byte from debug console; returns the byte on success, or -1 for failure.
+Note. This is the only SBI call that has a non-void return type.
+
+### Shutdown
+```C
+void sbi_shutdown(void)
+```
+Puts all the harts to shut down state from supervisor point of view.
```
palmer-dabbelt: This SBI call doesn't return.

riscv-sbi.md
aswaterman: That's a little misleading, since clearing sie.STIE just masks the timer interrupt, rather than clearing it.

Maybe something more explicit:

"If the supervisor wishes to clear the timer interrupt without scheduling the next timer event, it can either request a timer interrupt infinitely far into the future (i.e., (uint64_t)-1), or it can instead mask the timer interrupt by clearing sie.STIE."

riscv-sbi.md
atishp04: Fixed.

riscv-sbi.md
atishp04: Fixed already.

riscv-sbi.md
atishp04: Added.

atishp04:
Addressed all the feedbacks and updated some content as well. I have updated the same PR to maintain the history. Let me know if you want me to close this and create a new PR.

aswaterman:
Thank you, Atish!  Great start on this document.

PR #3, opened 2018-08-07
Typography, enhanced readability, refined ABI defs

sam-falvo:
Instead of using host-ambiguous types like unsigned long, prefer specific types instead, such as void *, uint8_t, etc.

sbi_console_putbyte(uint8_t) instead of (int) because anything outside of 0x00..0xFF is undefined behavior and almost certainly a bug.  Best to let the compiler detect this case; hurts no one.

Replace non-descript longs with asid_t to, again, aid in a compiler's or linter's ability to detect errors.  Yes, ASIDs are integers.  But they're specific kinds of integers.  The result of an arbitrary ASID + n is, except perhaps inside a unit test, almost certainly not what the programmer intended.

Attempt to fix anchors so that they're proper links per HTML specifications.

### DIFF
--- a/riscv-sbi.md
+++ b/riscv-sbi.md
@@ -109,7 +109,7 @@ rounded up to the next integer.
 If hart_mask is `NULL`, an IPI is sent to every hart in the system.
 
 ```C
-	void sbi_clear_ipi(void)
+void sbi_clear_ipi(void)
 ```
 Clears the pending IPIs if any. The IPI is cleared only in the hart for which
 this SBI call is invoked.
--- a/riscv-sbi.md
+++ b/riscv-sbi.md
@@ -2,14 +2,15 @@
 
 ## Table of Contents
 1. [Introduction](#Introduction)
-2. [SBI List](#sbi-list)
-3. [Description](#Description)
+1. [SBI List](#sbi-list)
+1. [Types](#Types)
+1. [Description](#Description)
 	* [Timer](#Timer)
 	* [Inter Processor Interrupt (IPI)](#IPI)
 	* [Memory Model](#memory-model)
 	* [Console](#Console)
 	* [Shutdown](#Shutdown)
-4. [Conclusion](#Conclusion)
+1. [Conclusion](#Conclusion)
 
 
 ## Copyright and license information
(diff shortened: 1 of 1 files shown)


aswaterman:
The SBI calls are already in use, so incompatible changes are a heavy lift.

void* and size_t are ABI-equivalent to unsigned long, so that’s technically OK (but sort of confusing, because the Linux bindings have different parameter types than this PR).

int vs. uint8_t isn’t compatible, since it moves the onus of truncation to the call site.

unsigned long is IMO a better type than void* for physical-address pointers, because it greatly reduces the odds that a user of the API will accidentally pass a virtual address.

I generally agree with using size_t when appropriate.

riscv-sbi.md
```diff
@@ -81,57 +106,64 @@ vector is represented as a sequence of unsigned longs whose length equals the
...
+                                size_t size,
+                                asid_t asid)
 ```
 //TODO
 
-### Console
+### Console<a name="Console" />
 ```C
-void sbi_console_putchar(int ch)
+void sbi_console_putbyte(uint8_t ch)
```
palmer-dabbelt: I think this is fine, as it doesn't break ABI compatibility (the ABI is the SBI number, not the function name).  We should update the Linux and BBL ports to the new name, do you want to submit patches?

palmer-dabbelt: Oh, sorry, Andrew's right -- I didn't notice the argument change.

sam-falvo:
Shucks; I wasn't aware that the SPI/SBI was already in use.  The sparsity of the document lead me to believe it was still in-flux.  I'll revert the types in a while.

aswaterman:
Yes, we're shamefully going back and documenting work we already did, and should've documented two years ago.

sam-falvo:
Are the SBI function numbers defined anywhere?  Or are those still fluid and changing?  I'm thinking I could incorporate that information into the document as well when I edit it next.

palmer-dabbelt:
@sam-falvo You should dig the numbers out of the source and put those in the doc, as that's the actual ABI (it looks just like a syscall).  Additionally, Andrew and I talked and I think "unsigned long" is a better mapping for the physical addresses.  Linux makes "phys_addr_t" an integral type, presumably to prevent people for accidentally leaking to/from virtual pointers.

We can just call it "phys_addr_t" as well, which brings up the issue of rv32 defining 34-bit physical addressees but systems probably wanting to also use 32-bit physical addresses for complexity reasons.

atishp04:
@sam-falvo : SBI function numbers
https://github.com/riscv/riscv-linux/blob/master/arch/riscv/include/asm/sbi.h#L19

atishp04:
Addressed some of the typo in a different PR. Closing the pull request.

Please raise another if you still find some issues.

PR #7, opened 2018-09-14
Some content updates

atishp04:
Please ignore the previous PR. For some reason, it did not diff correctly.

### DIFF
--- a/riscv-sbi.md
+++ b/riscv-sbi.md
@@ -60,8 +60,8 @@ environment. When executed in S-mode, it generates an environment-call-from-S-mo
 exception and performs no other operation. The required arguments are passed
 through registers *a0-a2* and the SBI call type is passed via register *a7*. Once
 the machine mode receives the trap, it identifies the type of SBI call from *a7*
-register and perform required operation. Any unsupported SBI call should return
-error i.e. -ENOSYS to indicate the supervisor mode that it is not supported.
+register and perform required operation. Any unsupported SBI call should return error code *-38*
+(which is the value of -ENOSYS in Linux) to indicate the supervisor mode that it is not supported.
 All SBI calls should be assumed to clobber a0 i.e. any return value will be passed
 through register *a0*. Individual SBI call signature and its purpose is described next.
 
--- a/riscv-sbi.md
+++ b/riscv-sbi.md
@@ -56,8 +56,9 @@ exception and performs no other operation. The required arguments are passed
 through registers *a0-a2* and the SBI call type is passed via register *a7*. Once
 the machine mode receives the trap, it identifies the type of SBI call from *a7*
 register and perform required operation. Any unsupported SBI call shoud return
-error i.e.ENOSYS to indicate the supervisor mode that it is not supported.
-Individual SBI call signature and its purpose is described next.
+error i.e. ENOSYS to indicate the supervisor mode that it is not supported.
+All SBI calls should be assumed to clobber a0 i.e. any return value will be passed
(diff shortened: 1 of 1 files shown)


riscv-sbi.md
```diff
@@ -55,20 +60,24 @@ environment. When executed in S-mode, it generates an environment-call-from-S-mo
 exception and performs no other operation. The required arguments are passed
 through registers *a0-a2* and the SBI call type is passed via register *a7*. Once
 the machine mode receives the trap, it identifies the type of SBI call from *a7*
-register and perform required operation. Individual SBI call signature and its
-purpose is described next.
+register and perform required operation. Any unsupported SBI call shoud return
+error i.e. ENOSYS to indicate the supervisor mode that it is not supported.
```
aswaterman: It's -ENOSYS, not ENOSYS, right?

atishp04: Yes. It should be -ENOSYS.

Apologies for the typo.

neuschaefer: I know Linux defines ENOSYS, but I think the SBI spec should not rely on such an explicit dependency, but rather define its ENOSYS explicitly

riscv-sbi.md
```diff
@@ -110,28 +119,32 @@ void sbi_remote_sfence_vma_asid(const unsigned long *hart_mask,
...
 Read a byte from debug console; returns the byte on success, or -1 for failure.
 Note. This is the only SBI call that has a non-void return type.
 
-### Shutdown
+```C
+void sbi_console_putchar(int ch)
+```
+Write data present in *ch* to debug console.
+
+Unlike `sbi_console_getbyte` , this SBI call **will block** if there
```
neuschaefer: It's still called _getchar above

atishp04: Thanks. I will fix that.

riscv-sbi.md
```diff
@@ -55,20 +60,24 @@ environment. When executed in S-mode, it generates an environment-call-from-S-mo
 exception and performs no other operation. The required arguments are passed
 through registers *a0-a2* and the SBI call type is passed via register *a7*. Once
 the machine mode receives the trap, it identifies the type of SBI call from *a7*
-register and perform required operation. Individual SBI call signature and its
-purpose is described next.
+register and perform required operation. Any unsupported SBI call shoud return
```
neuschaefer: s/shoud/should/

riscv-sbi.md
```diff
@@ -55,20 +60,24 @@ environment. When executed in S-mode, it generates an environment-call-from-S-mo
...
+error i.e. -ENOSYS to indicate the supervisor mode that it is not supported.
```
neuschaefer: As (I think) I've said before, I'd prefer if the value of `ENOSYS` was explicitly defined, rather than relying on implicit external definitions.

riscv-sbi.md
atishp04: Foxed.

riscv-sbi.md
atishp04: ENOSYS is a standard linux error code (38). Do you want me to just add the value as well ?

aswaterman: I agree with @neuschaefer -- this is not constant across operating systems, and the SBI should support more than Linux, so we should be explicit.  (Hurd defines it to be 78, and some Newlib ports use 88.)

I suggest writing something like "Any unsupported SBI call should return -38 (i.e., -ENOSYS)."

atishp04: Sounds good to me. Will update that.

PR #9, opened 2018-11-01
Change the return values

rminnich:
SBI needs to distinguish return values for the SBI calling sequence itself, and the values from the functions called. I think that we should reserve a0 for SBI errors, and a1 for the return value of functions themselves. Otherwise, as we've found in coreboot, there's not a good clean way to make console_putchar optionally compiled in; further, in using -38 for the error return from SBI, you will eternally mix up that one bit pattern with the bit pattern from all possible future SBI functions. Yuck.

This is how Unix originally worked: there were actually multiple return values from a system call (one of them being the PDP-11 carry bit, but still ...). It was only when we had to make the move to C, a single-return-value language, that we had to go with a single return value from system calls! And even in that case, there really are more than one: that's what errno is for: it's the "second return value."

Save yourself future pain and get this right now. Think of the call to SBI as you would an RPC, and allow that call to return errors; then allow the functions themselves to return an error distinct from possible SBI errors. You won't regret it.

### DIFF
--- a/riscv-sbi.md
+++ b/riscv-sbi.md
@@ -35,6 +35,9 @@ RISC-V. This provides a cleaner interface for the supervisor OS. Currently, SBI
 is currently implemented by Berkeley Boot Loader (BBL). The next section will
 describe different types of SBI calls defined for RISC-V.
 
+The SBI may be provided by the kernel, itself, or by firmware. That does not change the nature
+of the services provided.
+
 ## SBI List<a name="sbi-list" />
 There are mainly 5 types of SBI calls defined in the specification as per below
 table.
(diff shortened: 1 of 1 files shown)


rminnich:
so can you point me to it?

atishp04:
@rminnich 
http://lists.infradead.org/pipermail/linux-riscv/2018-October/001974.html

atishp04:
sounds to good to me. We can add one sbi_errno(as suggested by @wxjstz) that will provide the global SBI error stored in a0.

wxjstz:
@palmer-dabbelt !!! This looks great !!!

philhug:
> sounds to good to me. We can add one sbi_errno(as suggested by @wxjstz) that will provide the global SBI error stored in a0. If that's ok, I will update the proposal with this.

Do we still need sbi_errno() if we implement this proposal?
I'd rather not keep state if it's not needed.

@atishp04 could you upload your draft proposal as a PR? Might be easier to comment on.

atishp04:
> > @rminnich
> > http://lists.infradead.org/pipermail/linux-riscv/2018-October/001974.html
> 
> Hi Atish, thanks for linking to your proposal. As SBI changes affect much more than Linux, I think it would be worth at least CCing the sw-dev mailing list.

Yes. I should have done it in the first place. My bad.
I will update it with all the suggestions till now and send a v2 to both linux, sw-dev and raise a github PR as well for easy review.

atishp04:
> > sounds to good to me. We can add one sbi_errno(as suggested by @wxjstz) that will provide the global SBI error stored in a0. If that's ok, I will update the proposal with this.
> 
> Do we still need sbi_errno() if we implement this proposal?
> I'd rather not keep state if it's not needed.
> 
Yeah there are other concerns regarding the global state in the other thread (https://github.com/riscv/riscv-sbi-doc/issues/10) as well.

The other approach is to return error by reference or return a structure contain value & error. 
I like the structure approach as pointed by palmer. As per my understanding, psABI mandates that first value of the structure always goes into a0 & 2nd value goes into a1. So we are covered.

However, it means existing SBI call signature need to be modified. 
Given that, we are revamping SBI with versioning support, I guess that should be ok. 
Moreover, shall we make it mandatory return type for all SBI calls ?

> @atishp04 could you upload your draft proposal as a PR? Might be easier to comment on.
Yes. I will do that in proposal v2.

rminnich:
I just realized you are putting out sbi proposals on linux-riscv. This doesn't make a lot of sense to me. There are other kernels out there. I think you need to put them to sw-dev or this forum. 
I have a lot of things I'd like to see changed.

atishp04:
@rminnich : As I mentioned earlier, that was not intentional. I am planning to send a v2 to both sw-dev, & linux-riscv soon.  I will also create a github PR as well for easy review. Thanks for the feedback.

palmer-dabbelt:
We're doing the multi-return-value thing.

PR #12, opened 2018-11-10
proposal to extend SBI.

atishp04:
This PR is intended for discussion & feedback on the proposal.

### DIFF
--- a/riscv-sbi-prop.md
+++ b/riscv-sbi-prop.md
@@ -1,0 +1,290 @@
+
+## Introduction:
+This is a proposal to make SBI a flexible and extensible interface.
+It is based on the foundation policy of RISC-V i.e. modularity and
+openness. The proposal tries to introduces very few new mandatory SBI Functions,
+that are absolutely required to maintain backward compatibility. Everything else
+is optional so that it remains an open standard yet robust.
+
+The current RISC-V SBI only defines a few mandatory functions such as
+inter-processor interrupts (IPI) interface, reprogramming timer, serial
+console, and memory barrier instructions. The existing SBI documentation
+can be found here [1]. Many important functionalities such as power
+management/cpu-hotplug are not yet defined due to difficulties in
+accommodating modifications without breaking the backward compatibility
+with the current interface.
+
+The proposed design is inspired by Power State Coordination Interface (PSCI) from
+ARM world. However, it adds only two new mandatory SBI calls providing
+version information and supported Functions, unlike PSCI where a significant
+number of functions are mandatory. The version of the existing SBI will
+be defined as a minimum version(0.1) which will always be backward compatible.
+Similarly, any Linux kernel with a newer feature will fall back if an older
+version of SBI does not support the updated capabilities. Both the operating
+system and SEE can be implemented to be two way backward compatible.
+
+## SBI Calling Conventions (TODO)
+There are some suggestions[3] to distinguish between SBI function return value and
+SBI calling sequence return value. Here are the few possible discussed ideas.
+
+1. Return a single value, but values in the range {-511, -1} are considered
+errors, whereas return values in the range {0, 2^XLEN-512} are considered success.
+
+2. Every SBI function call may return the following structure
+```
+struct valerr {
+  long value;
+  long error;
+};
+```
+According to the RISC-V ABI spec[4], both a0 & a1 registers can be used for function
+return values.
+
+3. Introduce an extra argument that can return error by reference.
+```
+/* Returns the CPU that was started.
+ * @err indicates an error ID if one occurred, otherwise 0.
+ * @err can be NULL if no error detection is required.
+*/
+int start_cpu(int cpu_num,..., Error *err);
+```
+Both approaches (2 & 3) will work, but there will be two versions of existing SBI
+functions to maintain the compatibility.
+
+## SBI Functions:
+
+A SBI function is an individual feature that SBI interface provides to supervisor
+mode from machine mode. Each function ID is assigned as per below section.
+
+#### SBI Function ID numbering scheme:
+A Function Set is a group of SBI functions which collectively implement similar
+kind of feature/functionality. The function ID numbering scheme need to be scalable
+in the future. At the same time, function validation check should be as quick as possible.
+Keeping these two requirements in mind, a hybrid function ID scheme is proposed.
+
+SBI Function ID is u32 type.
+
+```
+31            24                        0
+-----------------------------------------
+|O|            |                        |
+-----------------------------------------
+Function Set   | Function Type          |
+```
+Bit[31]    =  ID overflow bit
+
+Bit[30:24] =  Function Set
+
+Bit[23:0]  =  Function Type within Function Set
+
+The Function set is Bits [31:24] describing both function set number and overflow
+bit. In the beginning, the overflow bit should be set to **zero** and the function
+Type per function set can be assigned by left shifting 1 bit at a time. The overflow
+bit can be set to one **only** when we need to allocate more than 24 functions per
+function set in future. Setting the overflow bit will indicate an integer function
+type scheme. Thus, there will more than enough function IDs available to use in the
+future. Refer appendix 1 for examples.
+
+Here are few Function Sets for SBI v0.2:
+
+| Function Set         | Value  | Description                                  |
+| -------------------  |:------:| :--------------------------------------------|
+| Base Functions       | 0x00   | Base Functions mandatory for any SBI version |
+| HART PM Functions    | 0x01   | Hart UP/Down/Suspend Functions for per-Hart power management|
+| System PM Functions  | 0x02   | System Shutdown/Reboot/Suspend for system-level power management|
+| Vendor Functions     | 0x7f   | Vendor specific Functions|
+
+N.B. There is a possibility that different vendors can choose to assign the same function
+numbers for different functionality. That's why vendor specific strings in
+Device Tree/ACPI or any other hardware description document can be used to verify
+if a specific Function belongs to the intended vendor or not.
+
+# SBI Function List in both SBI v0.2 and v0.1
+
+|Function Type              | Function Set      | ID(v0.2)     |ID (v0.1)  |
+|---------------------------| ------------------|:------------:|:---------:|
+| sbi_set_timer             | Base              | 0x00 000000  |0          |
+| sbi_console_putchar       | Base              | 0x00 000001  |1          |
+| sbi_console_getchar       | Base              | 0x00 000002  |2          |
+| sbi_clear_ipi             | Base              | 0x00 000004  |3          |
+| sbi_send_ipi              | Base              | 0x00 000008  |4          |
+| sbi_remote_fence_i        | Base              | 0x00 000010  |5          |
+| sbi_remote_sfence_vma     | Base              | 0x00 000020  |6          |
+| sbi_remote_sfence_vma_asid| Base              | 0x00 000040  |7          |
+| sbi_shutdown              | System PM         | 0x02 000000  |8          |
+| sbi_system_reset          | System PM         | 0x02 000001  |-          |
+| sbi_get_version           | Base              | 0x00 000080  |-          |
+| sbi_get_function_mask     | Base              | 0x00 000100  |-          |
+| sbi_get_function_count    | Base              | 0x00 000200  |-          |
+| sbi_hart_up               | Hart PM           | 0x01 000000  |-          |
+| sbi_hart_down             | Hart PM           | 0x01 000001  |-          |
+| sbi_hart_suspend          | Hart PM           | 0x01 000002  |-          |
+| sbi_hart_state            | Hart PM           | 0x01 000004  |-          |
+
+#### Function Description
+
+This section describes every newly introduced(in v0.2) function in details.
+Please refer to [1] for any v0.1 functions.
+
+```
+u32 sbi_get_version(void):
+```
+Returns the current SBI version implemented by the firmware.
+version: uint32: Bits[31:16] Major Version
+             Bits[15:0] Minor Version
+
+The existing SBI version can be 0.1. The proposed version will be at 0.2
+A different major version may indicate possible incompatible functions.
+A different minor version must be compatible with each other even if
+they have a higher number of features.
+
+```
+u32 sbi_get_function_mask(u32 ftype)
+```
+Given a function set type, it returns a bitmask of all functions IDs that are
+implemented for that function set.
+
+```
+u32 sbi_get_function_count(u32 ftype, u32 start_fid, unsigned long count):
+```
+
+This is a **reserved** function that should only be used if overflow bit in SBI
+function ID is set.
+
+Accepts a start_Function_id as an argument and returns if start_Function_id to
+(start_Function_id + count - 1) are supported or not.
+The Function numbering scheme is described in section 3.
+
+A count can help in minimizing the number of the M-mode traps by checking a range
+of SBI functions together.
+
+```
+int sbi_hart_up(unsigned long hartid, unsigned long start, unsigned
+long priv)
+```
+Brings up "hartid" either during initial boot or after a sbi_hart_down
+SBI call.
+
+"start" points to a runtime-specified address where a hart can enter
+into supervisor mode. This must be a physical address.
+
+"priv" is a private data that caller can use to pass information about
+execution context.
+
+Return the appropriate SBI error code.
+
+```
+int sbi_hart_suspend(u32 state, unsigned long resume_entry, unsigned
+long priv)
+```
+Suspends the calling hart to a particular power state. Suspended hart
+will automatically wake-up based on some wakeup events at resume_entry
+physical address.
+
+"priv" is a private data that caller can use to pass information about
+execution context. The SBI implementation must save a copy so that
+caller can reuse while restoring hart from suspend.
+
+Return the appropriate SBI error code.
+
+```
+int sbi_hart_down()
+```
+It powers off the hart and will be used in cpu-hotplug.
+Only individual hart can remove itself from supervisor mode. It can be
+moved to normal state only by sbi_hart_up function.
+
+Return the appropriate SBI error code.
+
+```
+u32 sbi_hart_state(unsigned long hartid)
+```
+Returns the RISCV_POWER_STATE for a specific hartid. This will help make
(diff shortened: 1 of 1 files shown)


riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+-----------------------------------------
+|O|            |                        |
+-----------------------------------------
+Function Set   | Function Type          |
+```
+Bit[31]    =  ID overflow bit
+
+Bit[30:24] =  Function Set
+
+Bit[23:0]  =  Function Type within Function Set
```
rminnich: I believe this is a mistake for several reasons, and we can talk if you are interested. You are building on the 1970s era model of a fixed set of well known entry points. This idea was created before we had things like a firmware device tree. You don't need to do what you are doing. You can require that vendors populate the FDT (the FDT they are required to create) with a set of function names and index pairs. The index is used by the handler in the SBI to call the function. The name is used to name the function. At boot, supervisor code can scan the FDT and populate its own table with the name/mapping pair. I've written lots of code like this and I can tell you it is easy.

riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+
```
rminnich: You can drop the rest of this section if you follow my advice.

riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+| HART PM Functions    | 0x01   | Hart UP/Down/Suspend Functions for per-Hart power management|
+| System PM Functions  | 0x02   | System Shutdown/Reboot/Suspend for system-level power management|
+| Vendor Functions     | 0x7f   | Vendor specific Functions|
+
+N.B. There is a possibility that different vendors can choose to assign the same function
+numbers for different functionality. That's why vendor specific strings in
+Device Tree/ACPI or any other hardware description document can be used to verify
+if a specific Function belongs to the intended vendor or not.
+
+# SBI Function List in both SBI v0.2 and v0.1
```
rminnich: you can drop this section if you follow my advice.

riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+```
+Returns the RISCV_POWER_STATE for a specific hartid. This will help make
+kexec like functionality more robust.
+
+```
+void sbi_system_reset(u32 reset_type)
+```
+Reset the entire system.
+
+## Return error code Table:
```
rminnich: don't do this. 

Error numbers should have high entropy, and not tie themselves to the
11 bits of the Linux kernel. 0, -1, and 1 are particularly bad
choices, as they can appear by mistake in many cases. Constants should
also follow this rule. 0-based enums are to be avoided. In coreboot,
we use values like 'DATA' and 'CODE', rather than 0 or 1, as
constants, and this makes for easy debugging and fault analysis, as
well as avoiding mistakes from using default uninitialized values like
0 or -1.

Debugging with just error numbers is frustrating and error prone. It
should be possible to read an error string after a call. The error
string should be stored in per-hart data areas. We accept that this is
an extra call, since errors should be rare, and the extra overhead is
acceptable. We put error strings in Akaros 5 years ago and thank ourselves
daily for that decision. As a 40 year firmware writer, I can tell you that error
strings in firmware are easy.

riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+generate both a firmware binary (to be executed directly in M mode) or a static library
+that can be used by different boot loaders. It will also help individual bootloaders
+to either work from M or S mode without a separate SBI implementation.
+
+## Conclusion
+This proposal is far from perfect and absolutely any suggestion is welcome.
+Obviously, there are many other functionalities that can be added to this proposal.
+However, I just wanted to start with something that is an incremental change at
+best to kick off the discussion. The aim here is to initiate a discussion that
+can lead to a robust SBI specification.
```
rminnich: there is not statement of principles with SBI. Is it intended to be like Alpha PALcode, as in the earlier design? is it intended to be a littler operating system, as Luke's requests imply (hence recreating 1970s BIOS?) How big do you want it to be? It seems here like you are building a little operating system. In my view that's not a good idea. In the LinuxBoot project, which we are doing at Google, Facebook, and others, we're trying to reduce the firmware activity to zero for very good security reasons. I view the growth of the SBI in these proposals as going the wrong direction.

A few other thoughts:
We should also acccommodate the changing language landscape: C is on the
way out, and languages like Rust and Go will dominate systems
programming.

The ABI should accommodate modern multiple-value return languages like
Rust and Go, and not be tied to the single return value restrictions
of C. 

Not all callers will be C and they won't always have C structure rules
for alignment and packing; even different C implementations have
different rules, standards notwithstanding. We need to consider that a
call to an SBI is like an RPC call, and we need to define language
independent formats for struct data. We got burned by this on Blue
Gene Q, where the struct format was not even compiler independent.

The set of services provided should be self describing and statically
enumerable via the FDT. We should not require a call to the SBI to
find out what services it provides.

SBI services should be separable, i.e. we should be able to replace
some without replacing the others. There should not be an overarching
version number for the entire set of firmware services on a single
implementation.

The table of services should be presented via the FDT as text, which
is endian- and alignment- independent. The text should be parseable to
a descriptor and an integer to be used as an SBI call number. While it
is desirable to have a small, contiguous range of call numbers, we
should not require it. This parsing need only be done once at boot.

The SBI should not include self-modifying code.

For any SBI service, it should be possible to statically enumerate
other services it depends on.

It should always be possible for a kernel to replace some or all of
the SBI with its own implementation.

riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+
+[1] http://infocenter.arm.com/help/topic/com.arm.doc.den0022d/Power_State_Coordination_Interface_PDD_v1_1_DEN0022D.pdf
+
+[2] https://github.com/riscv/riscv-sbi-doc/blob/master/riscv-sbi.md
+
+[3] https://github.com/riscv/riscv-sbi-doc/pull/9
+
+[4] https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md
+
+## Appendix
```
rminnich: The fact that you need to do this at all is a warning that you are walking into a trap.

riscv-sbi-prop.md
rminnich: in fact if you do this right, you can actually use the kernel's module loader to link to the sbi, which would be the simplest thing of all. Or you can use the coreboot relocatable module format. 

I'm sorry to be so hard on this idea but this (to me) is a very backwards looking idea on a very forward looking architecture.

olofj: Having a platform that can be self-probed without passing the right data to it is a benefit. It doesn't mean you can't _also_ come up with a binding that does higher-level descriptions of what is implemented, or a table with handles such that you don't have to probe. But having the API being capable of probing is beneficial, IMHO.

rminnich: It has been remarked that calls into M mode are expensive. This seems to be largely due
to the cost of the context switch. It is desirable, in my view, to be able to do a 
quick SBI call where no state is saved, but a register is loaded with a desired value,
as well as to be able to tell the SBI that it need not save all the state it saves now.

I'd like to propose an alternative calling model for the SBI, aimed at greater efficiency.
This assumes that we can move away from the hard-code SBI numbers model to my proposed
model in which SBI calling information is passed in the FDT, and S mode can create a
mapping of SBI function to call information once booted. This dynamic mapping
allows us to put some firmware-provided attributes into the call to improve efficiency.

To date, SBI call handlers, including the coreboot one, have taken the common approach
that kernels take: trap handler saves state and dispatches to a handler. The handler
then determines what to do based on the trap type. 

The problem is that for very simple register IO, this is a very expensive approach. It's
not currently possible, for example, to do a simple SBI that loads a result into A0,A1 and
returns.

I'd like to structure the SBI call number into two pieces:
-----------------------------
| Call #          | skip #             |
-----------------------------

The Call # will be an SBI call # as before, determined from the FDT.
The skip # will indicate how much
context saving can be skipped. A 0 will indicate none can be skipped;
a 1 will indicate, say,that R1 can be skipped, and so on.

The SBI will provide information about how much saving can be skipped; the caller
can choose to use that information as it wishes. The user might continue to request
that all state be saved; it might request that no state be saved, as it is using
caller save (as in Ken C and Go); or it might use the number provided by the FDT.

The Call # and skip # are not connected. An S-mode binary can call a function,but
save all the needed registers ahead of time, and M mode code might dirty many of
them. Hence
1:31
1:0
call the same SBI call, but S-mode asks for different levels of register saving to occur.

Similarly, the firmware can provide a function and indicate how many registers should
be saved:
MTIMER:0

means that the MTIMER function will dirty no registers and the kernel can, should it wish,
call it with any value between 0 and 31 and get the correct result.

Should the SBI indicate to the kernel that it should save many registers, the kernel
can do this by marking functions as caller-save via a compiler attribute,
or by using a larger skip #. There is more than one way to save context for an SBI call, 
either before the call, or by asking the SBI to do it.

In this way it is possible for the firmware to provide very efficient functions
where no state need be saved; or, for a kernel to arrange a call such that no state
need be saved.

Note that dispatch will now occur in 2 parts. The first, in assembly, will figure out
how much context to save, and save it; the second, in C, will figure out what function
to run.  This will require some rework of how coreboot implements SBI, but now is the
best time.

Note that the skip # of 0, meaning save everything, is chosen for compatibility with
current numbering schemes. Save everything could also be a larger number. 

Finally, I'd suggest the skip # field be 8 bits: more than is needed for now.

riscv-sbi-prop.md
atishp04: The return value from individual SBI calls still in discussion as I described in calling convention section. That's why I did not make any changes to it. My personal preference was to return a struct from discussion happened at https://github.com/riscv/riscv-sbi-doc/pull/9

Having error strings are definitely helpful. But that will lead to a hart local state. Since all of us want SBI implementation to be minimalistic, is it worth the extra management ?

riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Implementation
+Currently, SBI is implemented as a part of BBL. There is a different SBI implementation
+available in coreboot as well.
+
+Alternatively, a separate open BSD/MIT licensed SBI project can be created that
+can be used by anybody to avoid this kind of SBI fragmentation. This project can
+generate both a firmware binary (to be executed directly in M mode) or a static library
+that can be used by different boot loaders. It will also help individual bootloaders
+to either work from M or S mode without a separate SBI implementation.
```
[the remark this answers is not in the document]
atishp04: Agreed. I will remove it.

riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+
+This is a **reserved** function that should only be used if overflow bit in SBI
+function ID is set.
+
+Accepts a start_Function_id as an argument and returns if start_Function_id to
+(start_Function_id + count - 1) are supported or not.
+The Function numbering scheme is described in section 3.
+
+A count can help in minimizing the number of the M-mode traps by checking a range
+of SBI functions together.
```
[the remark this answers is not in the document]
atishp04: Hmm. Keeping both 32bit/64bit architecture, u32 should be function ID data type. Any objection to it ?

sbi_check_api() will lead to linear probing for each function which might be costly. I think you had a similar concern in the first version as well. If we allow it to return a bitmask we are limited by only 24 functions per function set only. That's the reason, this new function ID scheme is proposed. It is bit complicated for sure. Is probing individual functions ID is ok then (it was the scheme in first version)?

rminnich:
BTW, A good overview of how to pack data structures in a cpu- and endian- independent way can be found in http://9p.cat-v.org/documentation/rfc/, the draft 9p RFC. NOTE! I am NOT saying we should use 9p. I am using this document as a pointer to an example of ways to pack data structures so that they can be efficiently packed and unpacked. 

I know this sounds weird: we're calling M mode from S mode on the same machine, why do we need to unpack? And until I was working on a system with firmware written by IBM and a kernel written by someone else, compiled with different C compilers, both conformant to relevant standards, I would not have believed this could happen. But it does: firmware and kernel can disagree on how a struct is put in memory. Fact is, even if compilers *should* follow the rules, sometimes they don't, and sometimes rules change. 

At that point you end up having to write a structure unpacker in your kernel because the firmware layout of the structs, *even if they are declared the same way*, has a different layout in memory. And then you have to deal with versions of firmware, and how they might have compiled things.

We may start thinking we can keep the firmware so simple we don't have this struct problem; somebody always ends up adding them. 

Get ahead of this problem now by defining the packing format for structs using the style of notation defined in that RFC. That format has been the basis of working software defined in many languages, on many operating systems old and new, including Linux, for 30 years. It works.

riscv-sbi-prop.md
rminnich: I can tell you from experience that error strings, vs. just an errno, is a game changer. 

I don't see how, with the complexity of what you are proposing here, that you can avoid per-hart state in M mode. If you're going to do that, then you might as well have errstr.

riscv-sbi-prop.md
rminnich: Per Bruce Hoult's nice note today, this little bit of optimization is probably not what we'll do. We will need to rework but I think his idea is the best.

avpatel:
Like mentioned on LKML, the SBI spec should have minimum dependency on HW description language (FDT or ACPI) hence having SBI info in FDT is out of question.

IMHO, following are open questions (requiring immediate attention):
1. Should we return two values (error + data)? (My opinion, yes)
2. What should be good range of error codes? (My opinion 0-1024. Again it all depends on how we error code is returned)
3. How can we make sbi_check_api() better? (My opinion, current version is fine but it will be interesting to see newer approaches for SBI function discovery without using FDT/ACPI)

Regards,
Anup

asb:
Do you have a link to the LKML discussion where the use of FDT was ruled out?

atishp04:
@asb http://lists.infradead.org/pipermail/linux-riscv/2018-November/001995.html
@rminnich : Thanks a lot for all the feedback & apologies for the delay in response.

Dynamic discoverability is a good idea but doing via DT is the problem. We don't want SBI to be dependent on DT. We had some great discussions at RISC-V miniconf Plumbers about this as well. 

Here are some notes (FYI: These are only discussion notes. Not set in stone statements)
https://etherpad.openstack.org/p/RISC-V

At least everybody in that room seems to be on the same page. We will adopt some form of dynamic probing without DT support. The focus is to finalize the base specification first.

atishp04:
> BTW, A good overview of how to pack data structures in a cpu- and endian- independent way can be found in http://9p.cat-v.org/documentation/rfc/, the draft 9p RFC. NOTE! I am NOT saying we should use 9p. I am using this document as a pointer to an example of ways to pack data structures so that they can be efficiently packed and unpacked.
> 
> I know this sounds weird: we're calling M mode from S mode on the same machine, why do we need to unpack? And until I was working on a system with firmware written by IBM and a kernel written by someone else, compiled with different C compilers, both conformant to relevant standards, I would not have believed this could happen. But it does: firmware and kernel can disagree on how a struct is put in memory. Fact is, even if compilers _should_ follow the rules, sometimes they don't, and sometimes rules change.
> 
> At that point you end up having to write a structure unpacker in your kernel because the firmware layout of the structs, _even if they are declared the same way_, has a different layout in memory. And then you have to deal with versions of firmware, and how they might have compiled things.
> 
> We may start thinking we can keep the firmware so simple we don't have this struct problem; somebody always ends up adding them.
> 
> Get ahead of this problem now by defining the packing format for structs using the style of notation defined in that RFC. That format has been the basis of working software defined in many languages, on many operating systems old and new, including Linux, for 30 years. It works.

Thanks. I will take a look.

riscv-sbi-prop.md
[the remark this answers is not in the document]
atishp04: ok. A simple integer scheme may be sufficient.

riscv-sbi-prop.md
```diff
@@ -0,0 +1,290 @@
...
+3. Introduce an extra argument that can return error by reference.
+```
+/* Returns the CPU that was started.
+ * @err indicates an error ID if one occurred, otherwise 0.
+ * @err can be NULL if no error detection is required.
+*/
+int start_cpu(int cpu_num,..., Error *err);
+```
+Both approaches (2 & 3) will work, but there will be two versions of existing SBI
+functions to maintain the compatibility.
```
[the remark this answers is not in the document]
atishp04: ok. We can approach 2 seems to be a good choice then.

riscv-sbi-prop.md
atishp04: Agreed. Will remove this scheme.

riscv-sbi-prop.md
atishp04: @rminnich 
There is no doubt that SBI shouldn't be miniOS or too expansive. That's why Palmer suggested that we should start with only base SBI features and ratify them asap. Next, we can look into System/hart power management/vendor SBI extensions (can be worked in parallel by different group of people).
Any extension that need to be added should be only done if that's the only or best approach to do this. It should be avoided if an alternate sane approach exists.

> The set of services provided should be self describing and statically
> enumerable via the FDT. We should not require a call to the SBI to
> find out what services it provides.
> 

Please check the below answer about DT statements.

> SBI services should be separable, i.e. we should be able to replace
> some without replacing the others. There should not be an overarching
> version number for the entire set of firmware services on a single
> implementation.
> 

Not sure what do you mean. All the extensions except the base one will be optional. So firmware may choose not to implement any of that. We may have minor version number increments per extensions depending on if they take longer to ratify. Is that sufficient ?

> The SBI should not include self-modifying code.
> 
> For any SBI service, it should be possible to statically enumerate
> other services it depends on.
> 
> It should always be possible for a kernel to replace some or all of
> the SBI with its own implementation.

ACK. I will put those in a separate section "statement of principles".

rminnich:
This discussion seems to have died. Or is it occurring elsewhere?

atishp04:
> This discussion seems to have died. Or is it occurring elsewhere?
@rminnich 
We were waiting for an official riscv working group to restart the discussion and make it an official specification. The SBI talk at FOSDEM will provide more details. I guess you are planning to attend FOSDEM this year. I hope we can meet and discuss more details.

palmer-dabbelt:
I'm closing this because of the other v0.2 version.

PR #14, opened 2019-03-13
Start v0.2 of the SBI spec

palmer-dabbelt:
This is quite different.

Signed-off-by: Palmer Dabbelt <palmer@sifive.com>


riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+succeed.  The minor number of the SBI specification is encoded in the low 24
+bits, with the major number encoded in the next 7 bits.  Bit 32 must be 0 and
+is reserved for future expansion.
+
+```
+long sbi_get_sbi_impl_id(void);
+```
+Returns the current SBI implementation ID, which is different for every SBI
+implementation.  It is intended that this implementation ID allows software to
+probe
```
neuschaefer: Missing end of sentence.

~~How are implementation IDs assigned to implementations?~~ Perhaps mention that valid implementation IDs are listed below, and how new ones should be requested.

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+```
+Returns the current SBI implementation ID, which is different for every SBI
+implementation.  It is intended that this implementation ID allows software to
+probe 
+
+````
+long sbi_get_sbi_impl_version(void);
+````
+Returns the current SBI implementation version.  The encoding of this version
+number if specific to the SBI implementation.
```
neuschaefer: s/if specific/is specific/

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+optional modular extensions.
+
+### Versioning and History
+
+This document describes a draft of version 0.2 of the RISC-V SBI specification.
+
+#### Changes Since Version 0.1
+
+* The entire v0.1 SBI has been moved to the legacy extension, which is now an
+  optional extension.  This is technically a backwards-incompatible change,
```
neuschaefer: The comma leads to nowhere

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+Many SBI extensions also chose to encode an additional function ID in `a6`,
+a scheme similar to the `ioctl()` system call on many UNIX operating systems.
+This allows SBI extensions to encode multiple functions within the space of a
+single extension.
+
+In the name of compatibility, SBI extension IDs and SBI function IDs are
+encoded as signed 32-bit integers.  When passed in registers these follow the
+standard RISC-V calling convention rules.
+
+SBI functions generally return a pair of values in `a0` and `a1`, with `a1`
```
neuschaefer: "generally" → are there SBI functions that do not set a0 and a1?

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+|  SBI_ERR_INVALID_ADDRESS | -5     |
+
+
+## SBI Base Functionality, Extension ID 0x10
+
+The base of the supervisor binary interface is designed to be as small as
+possible.  As such, it only contains functionality for probing which SBI
+extensions are available and for querying the version of the SBI.  All
+functions in the base must be supported by all SBI implementations, so there
+are no error returns defined.
```
neuschaefer: Is a1 always set to SBI_SUCCESS by these functions, or do they leave a1 in some other/unspecified state?

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+```
+long sbi_get_sbi_spec_version(void);
```
neuschaefer: If a1 is set, perhaps specify the return type as `struct sbiret`?

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+| 1                 | OpenSBI                    |
+
+## Legacy SBI Extension, Extension IDs 0x00 through 0x0F
+
+The legacy SBI ignores the function ID field, instead being encoded as multiple
+extension IDs.  Each of these extension IDs must be probed for directly.  This
+extension is deprecated in favor of the following extensions:
+
+| Function Name             | Replacement Extension |
+|---------------------------|:---------------------:|
```
gurjeet: For consistency, use the Colons in header separators of all columns.

```suggestion
|:-------------------------:|:---------------------:|
```

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+````
+long sbi_get_{mvendorid,marchid,mimpid}(void);
+````
+Returns a value that is legal for the corresponding CSR.  0 is always a legal
+value for any of these CSRs.
+
+### Function IDs
+
+| Function Name                 | Function ID |
+|-------------------------------|:-----------:|
```
gurjeet: For consistency, use the Colons in header separators of all columns.

```suggestion
|:-----------------------------:|:-----------:|
```

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+| sbi_send_ipi              |                   N/A |
+| sbi_remote_fence_i        |                   N/A |
+| sbi_remote_sfence_vma     |                   N/A |
+| sbi_remote_sfence_vma_asid|                   N/A |
+| sbi_shutdown              |                   N/A |
+
+### Extension IDs
+
+| Function Name             | Extension ID |
+|---------------------------|:------------:|
```
gurjeet: For consistency, use the Colons in header separators of all columns.

```suggestion
|:-------------------------:|:------------:|
```

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
+# RISC-V Supervisor Binary Interface Specification
+
+## Copyright and license information
+
+This RISC-V SBI specification is
+
+[%hardbreaks]
+(C) 2019 Palmer Dabbelt <palmer@sifive.com>
+(C) 2019 Atish Parta <atish.parta@wdc.com>
```
gurjeet: I am not aware of the procedures adopted in RISC-V community, so pardon my ignorance.

I wonder if having individuals as copyright holders is the right thing to do. I assume, in case of a copyright claim the Foundation would have to get approval from the Copyright holders in order to respond and represent them.

/cc @kasanovic @rickoco

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+    struct sbiret {
+        long value;
+        long error;
+    };
+
+Standard SBI error codes are listed below
+
+| Error Type               | Value  |
+| -------------------------|:------:|
```
gurjeet:
```suggestion
|:-------------------------:|:------:|
```

For consistency, use the Colons in header separators of all columns.

riscv-sbi.adoc
```diff
@@ -0,0 +1,174 @@
...
+number if specific to the SBI implementation.
+
+````
+long sbi_probe_extension(long extension_id);
+````
+Returns 0 if the given extension ID is not available, or an extension-specific
+non-zero value if it is available.
+
+````
+long sbi_get_{mvendorid,marchid,mimpid}(void);
```
gurjeet:
```suggestion
long sbi_get_mvendorid(void);
long sbi_get_marchid(void);
long sbi_get_mimpid(void);
```

It's better to be more descriptive in documentation, than being terse.

riscv-sbi.adoc
atishp04: I think if we should mandate that every SBI function beyond 0.2 (including base functions) returns struct sbiret to avoid confusion for other future extensions. 

How about this ?

a0 - will contain only return values
a1 - will  be error from function or SBI calling sequence. The implementation may choose to use a error mapping between internal error and SBI error codes.

a0 value is valid only if a1 is SBI_SUCCESS (0). For any base SBI function, a1 may always be set to SBI_SUCCESS (0).

riscv-sbi.adoc
atishp04: Every function return type will be changed to struct sbiret if above suggested changes are incorporated to calling convention.

riscv-sbi.adoc
atishp04: Lets keep the legacy function descriptions in 0.2 legacy section as the next patch is removing the 0.1 document completely.

palmer-dabbelt:
I went through the comments and I'm just going to go ahead and merge this so we have a place to start.

PR #16, opened 2019-05-01
Switch to asciidoc format

mickflemm:
Since the doc has an .adoc extension it should follow asciidoc formatting rules, not mardkown (https://github.com/github/markup). I formatted this using the rules on http://asciidoc.org/userguide.html


palmer-dabbelt:
Thanks!

PR #20, opened 2019-06-12
Commit a first draft of PM extension

mickflemm:
This is a first draft of a power management extension, based on Atish's initial proposal.

Signed-off-by: Nick Kossifidis <mick@ics.forth.gr>

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -271,6 +271,61 @@ call doesn't return.
 | *RESERVED*                |    0x09-0x0F
 |===
 
+== Power Management Extension, Extension ID: 0x504D (PM)
+
+The Power Management Extension introduces a set of functions that allow the Supervisor to perform per-hart power management operations.
+
+[source, C]
+----
+struct sbiret sbi_pm_hart_up(unsigned long hartid, unsigned long start_addr, unsigned
+long priv)
+----
+
+Brings up *hartid* either during initial boot or after an sbi_hart_down/suspend
+SBI call.
+
+*start_addr* points to a runtime-specified physical address, where a hart can resume execution after its initialization/resume sequence. Before jumping to *start_addr*, the hart MUST configure PMP if present and switch to Supervisor mode.
+
+*priv* is a pointer to a opaque data segment that the caller can use to pass information about
+execution context.
+
+Returns an appropriate SBI error code on sbiret->error.
+
+[source, C]
+----
+struct sbiret sbi_pm_hart_down()
+----
+
+Powers off the calling hart. The hart can only be brought back up through the `sbi_hart_up()` call.
+
+Returns an appropriate SBI error code, through sbiret->error.
+
+[source, C]
+----
+struct sbiret sbi_pm_hart_suspend(u32 power_state, unsigned long resume_addr, unsigned
+long priv)
+----
+
+Suspends the calling hart to a particular *power_state*, which is a platform-specific value, outside the scope of this specification. Suspended hart may automatically wake-up based on some wakeup event or trough a call to `sbi_hart_up()`. After wake-up the hart will resume its operation on the physical address pointed out by *resume_addr* in case of a wakeup event, or on *start_addr* in case of `sbi_hart_up()`.
+
+*priv* is a private data the caller can use to pass information about execution context. The SBI implementation must save a copy so that caller can reuse while restoring hart from suspend.
+
+Returns an appropriate SBI error code, through sbiret->error.
+
+----
+struct sbiret sbi_pm_hart_status(unsigned long hartid)
+----
+
+Returns the current status of *hartid* on sbiret->value, or an error through sbiret->error. The possible status values are shown on the table below.
+
+[cols="<,>",options="header,compact"]
+|===
+| Return value | Description
+| 0 | Powered up & operational
+| 1 | Powered up but at reduced energy consumption
+| 2 | Powered off
+|===
+
 == Experimental SBI Extension Space, Extension IDs 0x0800000 through 0x08FFFFFF
 
 No management.


riscv-sbi.adoc
```diff
@@ -271,6 +271,61 @@ call doesn't return.
...
+
+Brings up *hartid* either during initial boot or after an sbi_hart_down/suspend
+SBI call.
+
+*start_addr* points to a runtime-specified physical address, where a hart can resume execution after its initialization/resume sequence. Before jumping to *start_addr*, the hart MUST configure PMP if present and switch to Supervisor mode.
+
+*priv* is a pointer to a opaque data segment that the caller can use to pass information about
+execution context.
+
+Returns an appropriate SBI error code on sbiret->error.
```
fintelia: nit: Shouldn't the `->` be a `.` instead? If not, beware that markdown will convert the ascii art arrow into a unicode one: "→"

riscv-sbi.adoc
```diff
@@ -271,6 +271,61 @@ call doesn't return.
...
+[source, C]
+----
+struct sbiret sbi_pm_hart_up(unsigned long hartid, unsigned long start_addr, unsigned
+long priv)
+----
+
+Brings up *hartid* either during initial boot or after an sbi_hart_down/suspend
+SBI call.
+
+*start_addr* points to a runtime-specified physical address, where a hart can resume execution after its initialization/resume sequence. Before jumping to *start_addr*, the hart MUST configure PMP if present and switch to Supervisor mode.
```
fintelia: This is a workable but somewhat awkward interface. Having the hart to start with paging disabled likely means that execution has to jump to a trampoline page, while not providing an opaque "context" argument means that it could be somewhat annoying for a new hart to learn its own hartid and/or why it was booted. An alternative would just be to make this function also take values for `satp` and `a0` on the booting hart.

riscv-sbi.adoc
atishp04: Thanks for pointing out. I will change it to sbiret.error

riscv-sbi.adoc
atishp04: I am not sure if I understood your point clearly. Paging is enabled when harts enter in Supervisor mode (i.e. Linux kernel). The purpose of this APIs is to bring up secondary harts one by one instead of all of them spinning on some global variable. a0 will always contain the hartid (set by run time firmware i.e. OpenSBI).

fintelia: My understanding is that with the PM extension things will work like this: First a SBI provider like OpenSBI will start up and initialize the system. Then it will set `a0=hartid` and `a1=&dtb` and jump into Linux on a single hart. Linux will then do some early bookkeeping and setup page tables on the primary hart. Once that is done, it will call sbi_pm_hart_up for each of the secondary harts. At this point Linux has already initialized its virtual memory subsystem, but each of the secondary harts still come up with paging disabled. They then need to run some boiler plate to initialize page tables before they can start doing useful work. Any subsequent time the hart powers down and is resumed again this same boiler plate code is needed again. This all feels to me very much like x86's decision to bring up secondary cores in 16-bit mode: the primary core already did a bunch of boiler plate to deal with this, why should secondary cores need to as well?)

The document also currently doesn't specify the state of any registers on the secondary harts after boot (other than saying `$pc = start_addr` and hinting that the satp register is reset to zero) so I assumed that this meant that all general purpose registers started with undefined values. Though from your reply it sounds like you have some other behavior in mind?

atishp04: > My understanding is that with the PM extension things will work like this: First a SBI provider like OpenSBI will start up and initialize the system. Then it will set `a0=hartid` and `a1=&dtb` and jump into Linux on a single hart. Linux will then do some early bookkeeping and setup page tables on the primary hart. Once that is done, it will call sbi_pm_hart_up for each of the secondary harts. At this point Linux has already initialized its virtual memory subsystem, but each of the secondary harts still come up with paging disabled. They then need to run some boiler plate to initialize page tables before they can start doing useful work.

Isn't that happening currently as well ? The only difference would be cpus not spinning in kernel and strict boot order in stead of random. 

 Any subsequent time the hart powers down and is resumed again this same boiler plate code is needed again. This all feels to me very much like x86's decision to bring up secondary cores in 16-bit mode: the primary core already did a bunch of effort to deal with this, why should secondary cores need to as well?)
> 
One of the other benefits of this also to support kexec in Linux. In case of an early crash in kernel, it gets tricky with spinning cpus. This also helps in unifying suspend/resume and boot path.

> The document also currently doesn't specify the state of any registers on the secondary harts after boot (other than saying `$pc = start_addr` and hinting that the satp register is reset to zero) so I assumed that this meant that all general purpose registers started with undefined values. 
Though from your reply it sounds like you have some other behavior in mind?

 Currently we reset all general purpose registers in OpenSBI except ra, a0, a1 and a2. There is a plan to document all register states upon boot in the platform specification. It is just that we have not reached to that stage yet.

All secondary cores will also continue to contain hartid in a0 as it is done currently in OpenSBI.
https://github.com/riscv/opensbi/blob/master/lib/sbi/sbi_hart.c#L280

fintelia: Thanks for your reply. I do think this document should say that secondary cores always contain their hartid in a0 after being booted/rebooted. That is a critical piece of information to program against this interface.

One possible tweak that I'd be curious to hear your thoughts on would be to have this function take one extra parameter which would be the value to use for satp when the hart booted. Passing zero would retain the currently proposed behavior while other values would bring up the hart with paging already enabled. This could make it easier to add SMP support to simpler kernels and would impose only a negligible burden on OpenSBI because it already has to do an IPI with start_addr in the payload, so an extra 4/8 bytes wouldn't really make much difference and actually setting the CSR is only a single instruction. Secondary harts could use this functionality at boot simply by doing a sbi_hart_down if they lost the hart lottery which would almost completely unify that path with the suspend/resume one.

The entry code for primary and secondary harts would probably no longer be all that similar, but the only reason they're even close right now in Linux is because they both spend so much effort setting up paging (I only count 7 other non-paging related instructions shared between the two paths before diverging to start_kernel/smp_callin).

atishp04: > Thanks for your reply. I do think this document should say that secondary cores always contain their hartid in a0 after being booted/rebooted. That is a critical piece of information to program against this interface.
> 
Yes. I will update the PR with that information.

> One possible tweak that I'd be curious to hear your thoughts on would be to have this function take one extra parameter which would be the value to use for satp when the hart booted. Passing zero would retain the currently proposed behavior while other values would bring up the hart with paging already enabled. This could make it easier to add SMP support to simpler kernels and would impose only a negligible burden on OpenSBI because it already has to do an IPI with start_addr in the payload, so an extra 4/8 bytes wouldn't really make much difference and actually setting the CSR is only a single instruction. Secondary harts could use this functionality at boot simply by doing a sbi_hart_down if they lost the hart lottery which would almost completely unify that path with the suspend/resume one.
> 
> The entry code for primary and secondary harts would probably no longer be all that similar, but the only reason they're even close right now in Linux is because they both spend so much effort setting up paging (I only count 7 other non-paging related instructions shared between the two paths before diverging to start_kernel/smp_callin).

atishp04: About your other suggestion: If I understand you correctly, this is what you want.
1. All harts jump to linux as it happens currently.
2. If a hart looses hart_lottery, it calls sbi_hart_down.
3. kernel brings up these hart using sbi_hart_up and uses satp value passed around to setup paging in kernel.

If that's the case, it doesn't have any performance benefit. It may simplify paging setup path, but you have to invoke an SBI call for every other non-booting hart. That's even more work for a hart and not scalable at all.

fintelia: I wasn't trying to suggest any changes for Linux specifically. The Linux kernel could be left as is and only use the satp=0 version in which case performance would be virtually unchanged from what it would be without this tweak.

Rather, for toy/prototype operating systems the complexity of the early boot interface is a bigger factor (it is a proportionally larger part of the code base) and I was hoping to simplify it for them. The satp register is uniquely annoying to configure from S-mode but trivial to set for M-mode before the supervisor starts running.

atishp04:
Closing this one as I have opened a updated one. #23

PR #22, opened 2019-08-08
Fix the my name and email address.

atishp04:


aswaterman:
Sorry, Atish, your name shall be wrong forever.

PR #23, opened 2019-08-22
1st draft of Hart hotplug SBI extension.

atishp04:
This draft is renamed from power management to Hart hotplug extension
and the function names are changed accordingly. It also adds more
description to each call and fixed some table issues.

Signed-off-by: Nick Kossifidis mick@ics.forth.gr

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -70,13 +70,14 @@ Standard SBI error codes are listed below
 
 [cols="<,>",options="header,compact"]
 |===
-|  Error Type              |Value
-|  SBI_SUCCESS             |  0
-|  SBI_ERR_FAILURE         | -1
-|  SBI_ERR_NOT_SUPPORTED   | -2
-|  SBI_ERR_INVALID_PARAM   | -3
-|  SBI_ERR_DENIED          | -4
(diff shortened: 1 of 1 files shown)


riscv-sbi.adoc
```diff
@@ -257,6 +258,85 @@ call doesn't return.
...
+* SBI_ERR_FAILED
+
+[source, C]
+----
+struct sbiret sbi_hart_remove()
+----
+
+Mark the calling hart unavailable to the supervisor mode software. The hart can only
+be made available through the `sbi_hart_add` call. This is a synchronous call and
+is not expected to return if succeeds.
```
avpatel: Please add a statement that a hart can only remove itself.

I know it is implicit from sbi_hart_remove() function prototype but we should have an explicit statement here to avoid confusion.

xnox:
During LPC 2019:

- start/stop are better names than add/remove
- this is not power-management, hence it's putting the cores into/out of reset state with/without instruction execution
- hart state management extension, is a better name. Because `hotplug` is confusing.
- lots of bikeshedding

atishp04:
Thanks for the feedback. Updated the PR #24 . A new PR is created due to branch name change.

PR #24, opened 2019-09-10
1st draft of Hart state management SBI extension.

atishp04:
This draft is renamed from power management to Hart state management
extension and the function names are changed accordingly. It also
adds more description to each call and fixed some table issues.

Signed-off-by: Nick Kossifidis mick@ics.forth.gr

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -70,13 +70,14 @@ Standard SBI error codes are listed below
 
 [cols="<,>",options="header,compact"]
 |===
-|  Error Type              |Value
-|  SBI_SUCCESS             |  0
-|  SBI_ERR_FAILURE         | -1
-|  SBI_ERR_NOT_SUPPORTED   | -2
-|  SBI_ERR_INVALID_PARAM   | -3
-|  SBI_ERR_DENIED          | -4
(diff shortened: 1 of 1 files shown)


atishp04:
Any comments ?

riscv-sbi.adoc
```diff
@@ -257,6 +258,93 @@ call doesn't return.
...
+
+Stop the calling hart by removing it from the list of harts that are currently running
+supervisor mode software. The stopped hart will no longer execute supervisor or user mode
+code, and will not respond to inbound supervisor or user IPIs.
+A hart can only stop itself and can be only started again via the `sbi_hart_start`
+call. This is a synchronous call and is not expected to return if succeeds.
+
+*Returns* following SBI error code through sbiret.error only if it fails.
+
+* SBI_ERR_FAILED
```
aswaterman: Why would this call fail, other than SBI_ERR_NOT_SUPPORTED?

atishp04: This is added for future compatibility. Currently, stop() will just call WFI. But we may have some actual hardware calls in future that will perform more than WFI. If those calls fail, machine mode can return SBI_ERR_FAILED.

PR #27, opened 2019-10-07
Replace required legacy extensions

atishp04:


atishp04:
As per our discussion during RISC-V summit, this PR should be merged and tagged as feature complete. 
Can we merge this PR ?

riscv-sbi.adoc
```diff
@@ -70,13 +70,44 @@ Standard SBI error codes are listed below
...
+|  SBI_ERR_INVALID_PARAM     | -3
+|  SBI_ERR_DENIED            | -4
+|  SBI_ERR_INVALID_ADDRESS   | -5
+|  SBI_ERR_ALREADY_AVAILABLE | -6
+|===
+
+Every SBI function should prefer `unsigned long` as the data type. It keeps the
+specification simple and easily adaptable for all RISC-V ISA types (i.e. RV32,
+RV64 and RV128). In case the data is defined as 32bit wide, the upper 32 bits
+should be considered as reserved bits unless otherwise specified.
```
palmer-dabbelt: I don't like this: 32-bit types are well defined by the ABI and we should respect that.  If there's a 32-bit type in a function signature then it's probably a 32-bit quantity, so there's no reason to pack other stuff into that register.

riscv-sbi.adoc
```diff
@@ -70,13 +70,44 @@ Standard SBI error codes are listed below
...
+
+Any function, requiring a hart mask, need to pass following two arguments.
+
+* `unsigned long hart_mask` is a scalar bit-vector containing hartids
+* `unsigned long hart_mask_base` is the starting hartid from which bit-vector
+   must be computed.
+
+In a single SBI function call, maximum number harts that can be set is always XLEN.
+If a lower privilege mode needs to pass information about more than XLEN harts, it
+should invoke multiple instances of the SBI function call.
```
palmer-dabbelt: We should allow something to say "all harts", maybe -1 in _base?

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
 | *RESERVED*                |             |    0x09-0x0F |
 |===
 
+
+== Timer Extension, Extension ID: 0x54494D45 (TIME)
```
palmer-dabbelt: I'm not sure purposefully leaving the important extensions sparsely mapped is a good idea.  IDK how much performance is going to be left on the table here, it just seems needless.

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
...
+This replaces legacy timer extension (0x00). It follows the new calling convention
+defined in v0.2.
+
+[source, C]
+----
+struct sbiret sbi_set_timer(uint64_t stime_value)
+----
+Programs the clock for next event after *stime_value* time. *stime_value* is in absolute
+time. This function must clear the pending timer interrupt bit as well.
```
palmer-dabbelt: This needs some rewriting.  If you move your PR over to a branch in this repo then I'm happy to do so, but I don't think I can do that on one you own.

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
...
+
+If the supervisor wishes to clear the timer interrupt without scheduling the next
+timer event, it can either request a timer interrupt infinitely far into the
+future (i.e., (uint64_t)-1), or it can instead mask the timer interrupt by
+clearing sie.STIE.
```
palmer-dabbelt: Those are two different things: one clears the interrupt enable, the other clears the interrupt pending.  This shouldn't mention STIE.

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
...
+
+== IPI Extension, Extension ID: 0x735049 (IPI)
+This extension replaces the legacy extension (0x03). All the functions follow
+the `hart_mask` defined in binary encoding section.
+
+[source, C]
+----
+struct sbiret sbi_send_ipi(const unsigned long hart_mask, unsigned long hart_mask_base)
```
palmer-dabbelt: The `const` here doesn't make any sense.

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
...
+----
+Send an inter-processor interrupt to all the harts defined in hart_mask.
+Interprocessor interrupts manifest at the receiving harts as the Supervisor software
```
palmer-dabbelt: "Supervisor" shouldn't be capitalized.

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
...
+interrupts.
+
+*Returns* following possible values via sbiret.
+[cols="<,>",options="header,compact"]
+|===
+| Error code                | Description
+| SBI_SUCCESS               | IPI was sent to all the targeted harts successfully.
+|===
+
+== RFENCE Extension, Extension ID: 0x52464E43 (RFNC)
```
palmer-dabbelt: After looking into some of the remote fence instruction stuff I really don't think we're ready to define an SBI extension here.  I'm specifically worried about all these fences being one-sided (ie, blocking), which probably isn't what you want for remote fences.  IIRC we'd decided at the Zurich workshop to stop speculatively defining SBI extensions until we had a corresponding ISA extension that would make them somehow better to implement in M-mode than in S-mode, and I think this extension is a good one to wait on.

Yes, that means than an SBI-0.2-clean Linux will need to handle these as IPIs, but I think that's actually a good thing: it avoids baking unknows into the ABI, which allows us to experiment more freely and ideally figure out the right way to do this more quickly.

riscv-sbi.adoc
aswaterman: However, more-privileged software shouldn't trust that less-privileged software has adhered to the ABI.  In particular, it shouldn't assume sub-XLEN types have been properly extended.  So I think the less-privileged software should be required to adhere to the ABI, as Palmer suggests (e.g. 32 bit values always sign-extended in 64-bit registers), but more-privileged software should assume it is passed as unsigned long, then cast it back to uint32 to do the extension.

riscv-sbi.adoc
avpatel: > After looking into some of the remote fence instruction stuff I really don't think we're ready to define an SBI extension here. I'm specifically worried about all these fences being one-sided (ie, blocking), which probably isn't what you want for remote fences.

The blocking nature of remote TLB flushes is as-per expectation of
Linux kernel and it is also true for IPI based remote TLB flushes.
In trying to fix Glibc craches with OpenSBI (six months back), we
realized that non-blocking remote TLB flushes with SBI calls (or with
IPIs) causes Linux kernel to crash because if we return before TLBs
on other HARTs are updated then other HARTs can potentially do invalid
access using stale TLB enteries. Due to this reason, on architecutures
(such as ARM64) with ISA support for remote TLB flush, the remote TLB
flush instructions are defined as blocking. The IPI based remote TLB
flushes for x86_64 is also blocking.
(Refer, https://github.com/torvalds/linux/blob/master/arch/x86/mm/tlb.c#L708)

> IIRC we'd decided at the Zurich workshop to stop speculatively defining SBI extensions until we had a corresponding ISA extension that would make them somehow better to implement in M-mode than in S-mode, and I think this extension is a good one to wait on.

IIRC, (at Zurich workshop 2019) we had decided to keep remote TLB flush
calls as optional (and not remove it) so that in-future an ISA extension
can be defined and preferred over SBI call.

The rational for above (as discussed at Zurich workshop 2019) is:

1. If we don't have SBI extension and ISA extension for remote TLB
flush then Guest/VM don't have any way for remote TLB flushes. A
software emulated MMIO CLINT for Guest/VM is not a viable option
because to inject IPI to N HARTs we need to do N MMIO writes where
each MMIO write will be trap-n-emulated. This will be very slow
and not at all scalable for any hypervisor.

2. In absence of S-mode CLINT and remote TLB flush ISA extension,
the IPI based remote TLB flush will be quite slow compared to SBI
calls because each IPI call will again go through SBI call. We had
proven this fact using benchmarks as well. In fact, this performance
difference was discussed in Linux kernel mailing list long time back
(Refer https://patchwork.kernel.org/cover/10872349/). The newer SBI
RFENCE calls are even better than SBI v0.1 RFENCE calls because we
are avoiding unpriv access so with these new calls we will get even
better performance compared to IPI based remote TLB flushes.

3. The SBI extension for remote TLB flush does not remove the
possibility of defining an ISA extension in-future. In fact,
in-future the Linux kernel will decide at runtime on method to
use for remote TLB flushes as follows:
 a) First check if an ISA extension for remote TLB flush is
    available. If available then use it otherwise try step b
 b) Next check if S-mode CLINT is available. If available then
    use IPI based remote TLB flush otherwise try step c
 c) Lastly check if SBI RFENCE extension is available. If
    available then use it otherwise try step d
 d) We don't have any method for remote TLB flush so crash.

> 
> Yes, that means than an SBI-0.2-clean Linux will need to handle these as IPIs, but I think that's actually a good thing: it avoids baking unknows into the ABI, which allows us to experiment more freely and ideally figure out the right way to do this more quickly.

Please see previous comments

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
...
+[source, C]
+----
+struct sbiret sbi_hart_stop()
+----
+
+Stop the calling hart by removing it from the list of harts that are currently running
+supervisor mode software. The stopped hart will no longer execute supervisor or user mode
+code, and will not respond to inbound supervisor or user IPIs.
+A hart can only stop itself and can be only started again via the `sbi_hart_start`
+call. This is a synchronous call and is not expected to return if succeeds.
```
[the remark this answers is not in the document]
atishp04: Will fix it.

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
...
+----
+
+*Returns* the current status of *hartid* in sbiret.value, or an error through
+sbiret.error. The possible status values are shown on the table below.
+
+[cols="<,,>",options="header,compact"]
+|===
+| Name                 		| Value | Description
+| AVAILABLE                     |   0   | Already available
+| NOT_AVAILABLE         	|   1   | Not available
```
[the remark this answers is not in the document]
atishp04: Yes. I realized this (STARTED/STOPPED) is better while implementing it in Linux and OpenSBI. I will update this.

riscv-sbi.adoc
```diff
@@ -257,6 +290,291 @@ call doesn't return.
...
+== Hart State Management Extension, Extension ID: 0x48534D (HSM)
+
+The Hart State Management Extension introduces a set of functions that allow the
+supervisor software to request higher privilege mode to start/stop harts running
+in supervisor mode.
+
+[source, C]
+----
+struct sbiret sbi_hart_start(unsigned long hartid, unsigned long start_addr, unsigned
+long priv)
```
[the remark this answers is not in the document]
atishp04: Fair enough. I did not mention it earlier as I though all register state upon boot will go into the base unix platform specification in booting requirement section. But I think there is no harm in clarifying that here as well.

riscv-sbi.adoc
atishp04: ok.

riscv-sbi.adoc
atishp04: Yeah. I will fix it.

riscv-sbi.adoc
atishp04: Sure. But I don't have write access to the repo so I can't create a branch in this repo.
Can you grant me write access or create a branch so that I can push it to that branch ? I am fine with the update in here as well.

riscv-sbi.adoc
atishp04: Currently, there are no performance impact as we are doing a direct comparison. It was kept sparse to make IDs assignment more logical (ASCII text). But I don't strongly feel about this. So I am happy to modify them to continuous if you prefer.

riscv-sbi.adoc
atishp04: sure. Currently, I am manually modifying the hart mask to online mask if hart_mask is zero.
But assigning hbase to -1 will do as well. It will make life simple in OpenSBI as well :).

riscv-sbi.adoc
atishp04: ok. I will update this.

JohnAZoidberg:
Is it intended that `sbi_clear_ipi` hasn't made it into the IPI Extension?

avpatel:
Yes, sbi_clear_ipi() is not required for S-mode because clearing an IPI in S-mode means clearing SSIP bit of SIP CSR which can be done directly by S-mode software.

Regards,
Anup

PR #30, opened 2020-02-26
Fix layout of register table for sbi_hart_start

JohnAZoidberg:
Previously the table was not rendered properly, resulting in cells being in the wrong column.


atishp04:
There was a similar patch merged as a part of PR #28. Closing this one.

PR #32, opened 2020-02-26
Fix list rendering in RFENCE Extension

JohnAZoidberg:
Without an empty line, the two items are not rendered as a list but put inline with the previous line.


atishp04:
Thanks for the fixes. You can send all the commits as a single PR as well :)

PR #36, opened 2020-03-06
Console functions deprecated, Function number listings for Timer & IPI extensions

konrad-schwarz:
added note that the console functions are to be deprecated
added the function number tables for the (single-function) timer and IPI extensions
minor editorial changes

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -187,9 +187,15 @@ value for any of these CSRs.
 
 == Legacy SBI Extension, Extension IDs 0x00 through 0x0F
 
-The legacy SBI ignores the function ID field, instead being encoded as multiple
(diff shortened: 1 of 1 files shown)


joel-porquet:
Sorry for reopening this. I've been going through as many webpages as possible in the past hour, and cannot find the rationale for deprecating the console functions from SBI. Will it now be expected that the supervisor takes care of its own console entirely, without any help from SBI? Or am I missing something? Thanks.

avpatel:
Yes, general expectations is that S-mode software will manage it's own console. The M-mode software will only use console for error reporting and debugging. Of course, the M-mode software can also have separate/dedicated console port based underlying RISCV platform.

The SBI v0.1 console usage is still available in Linux kernel and it will only be removed after all SBI implementations have moved to SBI v0.2.

Regards,
Anup

PR #39, opened 2020-03-31
Add system reset extension

avpatel:
Adds SBI v0.2 compliant system reboot extension. It defines two functions:
1. sbi_reboot - A system reboot call with reboot type as parameter
2. sbi_shutdown - A system shutdown/poweroff call

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -620,6 +620,85 @@ state of the hart at the time of return value verification.
 | sbi_hart_get_status      	|           2 |     0x48534D
 |===
 
+== System Reset Extension, Extension ID: 0x53525354 (SRST)
+
+The System Reset Extension provides a function that allow the supervisor
+software to request system-level reboot or shutdown. The term "system"
+refers to the world-view of supervisor software and the underlying SBI
+implementation could be machine mode firmware or hypervisor.
+
+=== Function: System reset (ID #0)
+
+[source, C]
+----
+struct sbiret sbi_system_reset(unsigned long reset_type, unsigned long reset_reason)
+----
+
+Reset the system based on provided `reset_type` and `reset_reason`. This is
+a synchronous call and does not return if it succeeds.
+
+The `reset_type` parameter is 32 bits wide and has the following possible
+values:
+
+[cols=",", options="header,autowidth"]
+|===
+| Value                   | Description
+| 0x00000000              | Shutdown
+| 0x00000001              | Cold reboot
+| 0x00000002              | Warm reboot
+| 0x00000003 - 0xEFFFFFFF | Reserved for future use
+| 0xF0000000 - 0xFFFFFFFF | Vendor or platform specific reset type
+| > 0xFFFFFFFF            | Reserved (and non-existent on RV32)
+|===
+
+The `reset_reason` is an optional parameter representing the reason for
+system reset. This parameter is 32 bits wide and has the following possible
+values:
+
+[cols=",", options="header,autowidth"]
+|===
+| Value                   | Description
+| 0x00000000              | No reason
+| 0x00000001              | System failure
+| 0x00000002 - 0xDFFFFFFF | Reserved for future use
+| 0xE0000000 - 0xEFFFFFFF | SBI implementation specific reset reason
+| 0xF0000000 - 0xFFFFFFFF | Vendor or platform specific reset reason
+| > 0xFFFFFFFF            | Reserved (and non-existent on RV32)
+|===
+
+When supervisor software is running natively, the SBI implementation is
+machine mode firmware. In this case, shutdown is equivalent to physical
+power down of the entire system and cold reboot is equivalent to physical
+power cycle of the entire system. Further, warm reboot is equivalent to
+a power cycle of main processor and parts of the system but not the entire
+system. For example, on a server class system with a BMC (board management
+controller), a warm reboot will not power cycle the BMC whereas a cold
+reboot will definitely power cycle the BMC.
+
+When supervisor software is running inside a virtual machine, the SBI
+implementation is a hypervisor. The shutdown, cold reboot and warm reboot
+will behave functionally the same as the native case but might not result
+in any physical power changes.
+
+*Returns* one of the following possible SBI error codes through sbiret.error
+upon failure.
+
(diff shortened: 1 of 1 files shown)


davidlt:
This seems to match `EFI_RESET_SYSTEM` from UEFI, least `EFI_RESET_TYPE` (`EfiResetCold`, `EfiResetWarm`, `EfiResetShutdown`, `EfiResetPlatformSpecific`) part. I would assume that UEFI implementation ends up using this extension once available.

Example:
https://github.com/u-boot/u-boot/blob/1297989804d66c8db92efbf9fa2e6cc76965a943/drivers/firmware/psci.c#L108

and we wouldn't need to have CPU specific resets in U-Boot code, like:
https://github.com/u-boot/u-boot/blob/master/arch/arm/cpu/armv8/fsl-layerscape/cpu.c#L1242

avpatel:
We did not take EFI reset call as reference for SBI SRBT extension but it is surprising that it turned out to be similar.

Just like ARM world, EFI reset call can be mapped to SBI SRBT call. This will certainly keep things simple for EFI implementation in U-Boot and EDK2.

Regards,
Anup

davidlt:
I also quickly skimmed PSCI specification. Here are a few notes:
- Why do we call this "reboot" instead of "reset", which seems to be the name for API for UEFI and PSCI?
- I really liked how PSCI described things to the details, i.e. `sbi_reboot` does not describe the scope or what is suppose to do (e.g. large server, virtualization, simple desktop system).
- It seems that PSCI for implementation defined reset type allow passing extra data (I think, they call it "cookie"). Is that widely used? (I have not searched code bases for it) If yes, shouldn't we do something similar?
- What happens to PMP protected memory on reset (and different reset types)? IIUC PSCI in a well defined situations memory should be overwritten to avoid attacks. I don't think that today PMP protects anything secret. It only protects firmware itself today IIRC. Should there be requirements for vendor or platmform specific reset type regarding this?
- Also PSCI for warm reset states that if memory is preserved and not trashed by firmware it should be communicated via ACPI / DT to the next stage.

riscv-sbi.adoc
```diff
@@ -615,6 +615,53 @@ state of the hart at the time of return value verification.
...
+
+[cols="<,>",options="header,compact"]
+|===
+| Value                 	| Description
+| 0x00000000			| Shutdown
+| 0x00000001			| Cold reboot
+| 0x00000002			| Warm reboot
+| 0x00000003 - 0xEFFFFFFF	| Reserved for future use
+| 0xF0000000 - 0xFFFFFFFF	| Vendor or platform specific reset type
+| 0x100000000 - 2^XELN-1	| Reserved and unusable on RV32
```
jrtc27, line 641:
```suggestion
| 0x100000000 - 2^XLEN-1	| Reserved (and non-existent on RV32)
```
The description was a bit confusing, even though it's obvious when you stop and think about it what it's meant to mean. Alternatively:
```suggestion
| > 0xFFFFFFFF	| Reserved (and non-existent on RV32)
```
might be a nicer way to express the LHS.

avpatel: This suggestions looks good to me. I will update PR accordingly.

avpatel:
Addressed comments from @jrtc27 

Added more description for shutdown, cold reboot, and warm reboot as suggested by @davidlt 

Regards,
Anup

jrtc27:
> I also suggest that the differences between reset types be explicitly defined. For example, is
> 
> ```
> hang:
>     wfi
>     j hang
> ```
> 
> a valid implementation of shutdown?

I would say no, as it's not doing any power control. Better to let it fail and then if the caller would still like to hang then it can issue the WFI loop itself, since that's not platform-specific and thus SBI does not need to be involved.

> Is
> 
> ```
> reboot:
>     la a0, reset
>     j send_ipi_all
> 
> reset:
>     csrw mstatus, zero
>     li t0, 0xC3FFFFFF
>     csrs misa, t0
>     j <some address>
> ```
> 
> a valid implementation of warm reboot?

No, the next stage needs to be reloaded otherwise any data segments will not have their initial state (and zero-filled memory may have non-zero data left in it, though some OSes don't trust their loader and will manually zero it out in their entry point).

> More importantly, is either example a valid implementation of the other reset type? I don't think they should be, but this is left unspecified at the moment.

I agree; the first is clearly not a reset, and the second is clearly not a shutdown.

avpatel:
> SBI_ERR_NOT_SUPPORTED

Returning SBI_ERR_NOT_SUPPORTED for unsupported reset_types sounds good to me.

We can't have the suggested code snippets about SHUTDOWN and REBOOT implementation using WFI because SBI calls are also provided by hypervisors (KVM, Xvisor, etc). The best we can do is to functionally describe what is expected from various RESET TYPEs. We already have a brief description about SHUTDOWN, COLD REBOOT, and WARM REBOOT so please let me know if we need to improve that.

Regards,
Anup

avpatel:
> Lastly, there is an implied fourth reset option. A "warm shutdown" could power off, reset, or clock gate the "main processor and parts of the system but not the entire system." Upon a wakeup event from any of the powered peripherals, the system would behave as if it had just entered a warm reset. Should this be supported? Many existing processors have a "deep sleep" mode like this where nothing is powered except the RTC or some GPIOs.

You idea about "warm shutdown" sounds lot like "system suspend". We would be having separate SBI call for "system suspend" because it might involve additional parameter. Let's keep "system suspend" as separate topic not related to SBI SRST extension.

Regards,
Anup

riscv-sbi.adoc
```diff
@@ -618,6 +618,66 @@ state of the hart at the time of return value verification.
...
+
+When supervisor software is running natively, the SBI implementation
+is a machine mode firmware. In this case, shutdown is equivalent to
+physical power down of the entire system and cold reboot is equivalent
+to physical power cycle of the entire system. Further, the warm reboot
+is equivalent to power cycle of main processor and parts of the system
+but not the entire system. For example, on server class system with a
+board management controller (BMC), a warm reboot will not power cycle
+the BMC whereas a cold reboot will definetly power cycle the BMC.
+
```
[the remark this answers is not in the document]
avpatel: Okay, will update.

atishp04:
As it is already reviewed in the platform mailing list, merging it here.

Issue #40, opened 2020-04-01
Function to retrieve the Machine ISA CSR ``misa'' missing 


konrad-schwarz:
I would expect a function `struct sbiret sbi_get_misa (void);` to be provided.

The Privileged Architecture Spec makes a clear case for this:

> We require that lower privilege levels execute environment calls
> instead of reading CPU registers to determine features available
> at each privilege level. This enables virtualization layers
> to alter the ISA observed at any level, and supports a much richer
> command interface without burdening hardware designs.

JohnAZoidberg:
Can be closed, since #41 has been closed.

PR #41, opened 2020-04-01
Added sbi_get_misa()

konrad-schwarz:
See issue #40.

To restate, the Privileged ISA Spec says it all:
> We require that lower privilege levels execute environment calls instead of reading CPU
> registers to determine features available at each privilege level. This enables virtualization layers
> to alter the ISA observed at any level, and supports a much richer command interface without
> burdening hardware designs.

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -153,6 +153,7 @@ non-zero value if it is available.
 
 [source, C]
 ----
+struct sbiret sbi_get_misa(void);
 struct sbiret sbi_get_mvendorid(void);
 struct sbiret sbi_get_marchid(void);
 struct sbiret sbi_get_mimpid(void);
@@ -172,6 +173,7 @@ value for any of these CSRs.
 | sbi_get_mvendorid             |           4 |         0x10
 | sbi_get_marchid               |           5 |         0x10
 | sbi_get_mimpid                |           6 |         0x10
+| sbi_get_misa                  |           7 |         0x10
 |===
 
 === SBI Implementation IDs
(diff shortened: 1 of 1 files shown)


avpatel:
The RISC-V feature string is much more richer and future proof compared to a single unsigned long misa CSR. Due to this reason we are parsing RISC-V feature string from device tree in various S-mode software (e.g. Linux, U-Boot, EDK2, Xvisor, KVM RISC-V, and many more).

In future, ACPI will also include RISC-V feature string which S-mode software can use.

I don't see why we need this SBI call.

Regards,
Anup

konrad-schwarz:
I didn't realize that you were using the device tree for this.  Perhaps some commentary in the SBI spec would reduce confusion here.

I will note that (especially for small systems such as a booloader) it is much simpler to test bits in a binary register such as `misa` than to extract (or possibly generate/modify) them from a string-valued feature string; this becomes more true the older an architecture gets; cf. the x86 CPUID command, which is (mostly) binary and not string based, or the ARM CPUID architecture,, which also uses binary encoding.

Also, you argument applies just as well to the three static CSRs that are exposed by the SBI: `machid`, `mimpid`, `mvendorid`:  these could just as well be transported by the device tree.
I find leaving out `misa` inconsistent.

avpatel:
The RISC-V feature string does not include machid, mimpid, mvendorid so that's why we have these calls from SBI v0.2 onwards.

Regards,
Anup

konrad-schwarz:
My suggestion above would be to add these items to the device tree RISC-V binding, which is decidedly more flexible than the SBI API.

So in addition to the existing `riscv,isa` string valued node, you would also have `rsicv,marchid`, `riscv,mvendorid`, `riscv,mimpid`, and (posssibly) `riscv,misa` binary nodes in the device tree.

atishp04:
Closing this PR as I commented on #48

PR #43, opened 2020-04-09
Firmware Code Base SBI Extension space

changab:
Reserve extension space for firmware code base specific SBI extensions.

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -622,3 +622,10 @@ No management.
 == Vendor-Specific SBI Extension Space, Extension Ids 0x09000000 through 0x09FFFFFF
 
 Low bits from `mvendorid`.
+
+== Firmware Code Base Specific SBI Extension Space, Extension Ids 0x0A000000 through 0x0AFFFFFF
+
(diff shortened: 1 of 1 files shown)


riscv-sbi.adoc
```diff
@@ -615,7 +615,13 @@ state of the hart at the time of return value verification.
...
 |===
 
-== Experimental SBI Extension Space, Extension IDs 0x0800000 through 0x08FFFFFF
+== Firmware Code Base SBI Extension Space, Extension IDs 0x0800000 through 0x08FFFFF
+
+Firmware code base (e.g. edk2 or uboot) may define its own SBI extension for the firmware
+drivers built on top of it. This SBI extension space is reserved for the firmware code base 
+specific SBI Extension.
+
+== Experimental SBI Extension Space, Extension IDs 0x0900000 through 0x08FFFFFF
```
JohnAZoidberg, line 624: The beginning ID is after the end ID.
I think you mean:

```suggestion
== Experimental SBI Extension Space, Extension IDs 0x0900000 through 0x09FFFFFF
```

changab: Oh! yes. you are right! I will change it later. Thanks

PR #50, opened 2020-05-17
Add README

JohnAZoidberg:
It's nice to have a README to explain the context.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,0 +1,17 @@
+= RISC-V SBI specification
+
+SBI (Supervisor Binary Interface) is an interface between the Supervisor
+Execution Environment (SEE) and the supervisor. It allows the supervisor to
+execute some privileged operations by using the ecall instruction. Examples of
+SEE and supervisor are: M-Mode and S-Mode on Unix-class platforms, where SBI is
+the only interface between them, as well as the Hypervisor extended-Supervisor
+(HS) and Virtualized Supervisor (VS).
+
+Discussion of SBI occurs on the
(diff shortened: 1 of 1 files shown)


changab:
I have no problem with adding README. How about others?

fintelia:
I'm in favor of adding a README. One point on the content: it should be clarified that SBI is only the interface between M-mode and S-mode for the Unix-class platform. It also might be good to spell out that SBI stands for "Supervisor binary interface".

avpatel:
Yes, a top-level README stating purpose of SBI is good to have. 

BTW, SBI the interface is between:
1) M-mode and S-mode
2) HS-mode and VS-mode

Regards,
Anup

JohnAZoidberg:
Updated to address your feedback.

README.adoc
```diff
@@ -0,0 +1,17 @@
+= RISC-V SBI specification
+
+SBI (Supervisor Binary Interface) is the only interface between M-Mode and
+S-Mode on Unix-class platforms. It allows S-Mode to execute some privileged
+operations by using the ecall mechanism.
```
atishp04, line 5: I think it's better to write in a generic way:
SBI (Supervisor Binary Interface) is an interface between Supervisor Execution Environment (SEE) and the supervisor. The M-Mode/S-mode and HS/VS mode can be described as an example to this.

atishp04:
@JohnAZoidberg Can you also add that the patches should be sent to the mailing list in order to be merged? Reviews on Github PRs are still fine. But, we will try to make sure that every patch lands in the mailing list before merging.

PR #57, opened 2020-09-02
Add SBI implementation ID for Diosix

diodesign:
Diosix is a bare-metal open-source hypervisor written in Rust for RISC-V systems that implements the RISC-V SBI to communicate with supervisor kernels. As such, it needs an implementation ID, which I request here. Diosix's source code and documentation can be found here: https://github.com/diodesign/diosix

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -183,6 +183,7 @@ value for any of these CSRs.
 | 1                 | OpenSBI
 | 2                 | Xvisor
 | 3                 | KVM
+| 4                 | Diosix
 |===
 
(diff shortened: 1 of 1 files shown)


atishp04:
@diodesign 
Can you please send an email to the unix platform mailing list as that is the preferred approach for discussing technical details ? Here are the details
https://github.com/riscv/riscv-sbi-doc/blob/master/README.adoc

This patch also needs to be revised to update the implementation id to 5 as there is another PR(#61 ) which also claimed implementation id 4.

diodesign:
OK, I'll take it to the mailing list, thanks.

PR #63, opened 2020-11-18
Top-level makefile and other cleanups

avpatel:
Added top-level makefile to build PDF/HTML documents and lots of other cleanups.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,28 @@
+#
+# Build usable documents
+#
+
+ASCIIDOCTOR = asciidoctor
+TARGETS = riscv-sbi.pdf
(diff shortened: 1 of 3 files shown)


atishp04:
You can revise the PR once only after it is reviewed in the mailing list to save the extra hassle.

avpatel:
I don't mind keeping PR updated. In fact, updating PR is relatively easy compared to sending patch emails.

Regards,
Anup

ahs3:
> You can revise the PR once only after it is reviewed in the mailing list to save the extra hassle.

Yeah, exactly.  Don't make extra work for yourself if you don't need to.

atishp04 (approved):

PR #67, opened 2021-03-09
Handling of unsupported EIDs and FIDs

xypron:
Up to now the SBI specification does not define how unsupported EIDs and
FIDs shall be handled. Require returning error code SBI_ERR_NOT_SUPPORTED.

Signed-off-by: Heinrich Schuchardt <xypron.glpk@gmx.de>


atishp04 (approved):
Usually, it is recommended to send the patches to the mailing list before merging it. However, this is a very minor clarification improvement. I am okay with merging this directly.

PR #69, opened 2021-04-29
SRST documentation improvment and table captions

avpatel:


atishp04 (approved):

atishp04:
Merging it as it is reviewed in the mailing list already.

PR #71, opened 2021-05-17
Fix HART get status extension typo

vpelletier:
Looks like a trivial typo, as this function does not take a `start_addr` argument.


avpatel:
Looks good to me

@atishp04 can you please take this PR

Regards,
Anup

atishp04 (approved):
Since this is a trivial change, merging it directly instead of waiting for reviews in the mailing list.

PR #73, opened 2021-06-07
Add performance monitoring unit extension

avpatel:
This patch adds SBI performance monitoring unit (PMU) extension which
allows S-mode (or VS-mode) software to configure hardware (or firmware)
performance counters with help of M-mode (or HS-mode) software.


atishp04 (approved):
LGTM.

PR #74, opened 2021-06-13
Some small changes

palmer-dabbelt:
These are just some small changes I came up with when reading through the spec.  None of them are meant to change the meaning of the spec (at least, as I interpreted it) but there were some vague bits I thought it'd be best to clarify.  I didn't make it all the way through the spec, but I figured it'd be easier to just post this to start.


avpatel:
Thanks Palmer. All your changes look good to me.

@atishp04 please check this at your end.

Regards,
Anup

atishp04 (approved):

atishp04:
Looks good to me.

PR #77, opened 2021-07-28
[PATCH 1/1] Proposal to add Coffer to the SBI impl IDs list

jwnhy:
Sorry, the format of the patch in the maillist seems incorrect.

So I opens a PR here.

The discussion link is [here](https://lists.riscv.org/g/tech-unixplatformspec/topic/patch_1_1_proposal_to_add/84507846?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,84507846).

Thanks again.


atishp04:
Thanks.

Issue #84, opened 2022-01-11
Update to RISC-V Template


jjscheel:
Need to update the same look and feel as the other RISC-V document.  This is accomplished by the following steps: 

1. Utilize GitHub submodules to include a docs-resources project.
2. Updates to the README document to explain how to clone with submodules
3. Creation of an image/ directory with a riscv-images symlink into the docs-resources
4. Updates to the Makefile to use the RISC-V fonts and themes when building
5. Removal of the old theme file (riscv-sbi-theme.yaml).
6. Addition of necessary Asciidoctor variables to the main adoc file (riscv-sbi.adoc).
7. Updating the document level to 1.0-rc2 and adding change log entry (riscv-sbi.adoc)

atishp04:
Merged in PR #89. Closing the issue

PR #86, opened 2022-01-12
Convert project to RISC-V template (docs-resources)

jjscheel:
More details in #84 

Signed-off-by: Jeff Scheel <jeff@riscv.org>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "docs-resources"]
+	path = docs-resources
+	url = https://github.com/riscv/docs-resources.git
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 #
 
 ASCIIDOCTOR = asciidoctor
+ASCIIDOCTOR_PDF = $(ASCIIDOCTOR)-pdf
 DITAA = ditaa
 IMAGES = riscv-sbi-intro1.png
 IMAGES += riscv-sbi-intro2.png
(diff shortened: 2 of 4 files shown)


atishp04:
This doesn't build without
sudo gem install pygments.rb

It should be added to the README if it is required.

jjscheel:
Per this comment:

> This doesn't build without
> sudo gem install pygments.rb
>
> Probably, make install-deb/rpms should be updated accordingly.

I'm not sure I am the best person make the updates to the Makefile because I don't have an environment that fails and I don't have the background on what environments are being supported.  

Can you accept this PR and then make the updates to Makefile to meet your requirements?  While I fully acknowledge this feels like the PR is incomplete, but hopefully I've saved you the time learning the template environment.

BTW, I can't get `make clean; make` to fail from my Linux command line.  It works for me so I'm suspicious it's tied to the dependencies problem or perhaps I don't understand the context in which it's being executed.

atishp04:
You were right. I did not do a proper git submodule update which lead to the errors I was seeing.

jjscheel:
I should add those comments to the SBI README.  Should I re-submit the PR?

atishp04:
That would be great.

jjscheel:
@atishp04, I think I re-submitted the PR correctly.  It does now include 2 new sections in the README for Dependencies and Cloning the project.

Please let me know if you see any problems.

atishp04:
I have opened another PR #89 that folded this PR with some minor edits about the document state. Closing this one.

PR #87, opened 2022-01-20
Explicitly mention the integer data model

axel-h:


atishp04 (commented):
As per this thread, the current version will only ratified for RV64. Any reference to RV32 will be removed. 

Do you still think that we need the integer data model in that case ?

axel-h:
I'd like to add the URL for the Spec at least to have some reference. Mentioning the integer model is a helpful context, given `long` has no fixed meaning otherwise. Assuming the interface headers are supposed to be reused by the calling OS, using a wrapper type like `sbi_word` might be worth considering

atishp04:
Closing this one as this issue is resolved through #88

PR #89, opened 2022-01-27
public review comments & style change

atishp04:


axel-h:
Why did you drop the link to the RISC-V ELF psABI PDF?

atishp04:
because we are going to drop any reference to psABI soon. Check the discussions in #35

PR #90, opened 2022-02-08
Update the calling convention

atishp04:
SBI specification calling convention can not refer to the syscall or
psABI calling convention because of the current implementation
FreeBSD/Linux precludes that.

Moreover, all the major SBI implementations  KVM, Xvisor) saves
all the registers (except a0 & a1) anyways.

Update the calling convention to reflect that.

Signed-off-by: Atish Patra <atishp@rivosinc.com>

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -130,26 +130,20 @@ discovery methods (i.e. Device Tree or ACPI) for that.
 == Binary Encoding
 
 All SBI functions share a single binary encoding, which facilitates the mixing
-of SBI extensions. This binary encoding matches the RISC-V Linux syscall ABI,
-which itself is based on the calling convention defined in the RISC-V ELF
-psABI. In other words, SBI calls are exactly the same as standard RISC-V
-function calls except that:
+of SBI extensions. The SBI specification follows the below calling convention.
 
-* An `ECALL` is used as the control transfer instruction instead of a `CALL`
-  instruction.
+* An `ECALL` is used as the control transfer instruction between the supervisor
+  and the SEE.
+
 * `a7` encodes the SBI extension ID (*EID*),
-  which matches how the system call ID is encoded in Linux system call ABI.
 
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

riscv-sbi.adoc
```diff
@@ -160,6 +154,10 @@ returning an error code. This is analogous to returning the C structure
     };
 ----
 
+In the name of compatibility, SBI extension IDs (*EIDs*) and SBI function IDs
+(*FIDs*) are encoded as signed 32-bit integers. When passed in registers these
```
axel-h, line 158: - say "should" instead of "are"? New EIDs/FIDs may exceed this range and older software will never use them?

riscv-sbi.adoc
```diff
@@ -130,26 +130,20 @@ discovery methods (i.e. Device Tree or ACPI) for that.
...
-in `a6`, a scheme similar to the `ioctl()` system call on many UNIX operating
-systems. This allows SBI extensions to encode multiple functions within the
-space of a single extension.
+* `a6` encodes the SBI function ID (*FID*) for a given extension ID encoded
+   in `a7` for any SBI extension defined in or after SBI v0.2.
 
-In the name of compatibility, SBI extension IDs (*EIDs*) and SBI function IDs
-(*FIDs*) are encoded as signed 32-bit integers. When passed in registers these
-follow the standard RISC-V calling convention rules.
+* All registers except `a0` & `a1` must be preserved across an SBI call by the
```
axel-h, line 143: I assume there has been a decision that no extensions will ever be allowed to return additional data via a2-a7. Any such use case must span multiple SBI calls then or use memory then? Giving some rationale in a foot note might be helpful to clarify why these tight restriction are choose. This could link to GitHub/Workgroup discussion, but preferably the essence is given in the spec.

atishp04: RISC-V only allows a0/a1 to be hold return values.

PR #91, opened 2022-02-17
Improve documentation of ABI followed by legacy SBI extensions

avpatel:
The legacy SBI extensions (i.e. SBI v0.1) follow a slightly different
ABI where nothing is returned in a1 register and error code is returned
in a0 register. We need to document this fact correctly in the Legacy
extensions chapter.

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -330,9 +330,19 @@ value for this CSR.
 
 == Legacy Extensions (EIDs #0x00 - #0x0F)
 
-The legacy SBI extensions ignores the SBI function ID field, instead being
-encoded as multiple SBI extension IDs. Each of these extension IDs must be
-probed for directly.
+The legacy SBI extensions follow a slightly different calling convention
+as compared to the SBI v0.2 (or higher) specification where:
+
+* The SBI function ID field in `a6` register is ignored because these are
+  encoded as multiple SBI extension IDs.
+* Nothing is returned in `a1` register.
+* All registers except `a0` must be preserved across an SBI call by the
+  callee.
+* The value returned in `a0` register is SBI legacy extension specific.
+
+The page and access faults taken by the SBI implementation while accessing
+memory on behalf of the supervisor are redirected back to the supervisor
+with `sepc` CSR pointing to the faulting `ECALL` instruction.
 
 The legacy SBI extensions is deprecated in favor of the other extensions
 listed below. The legacy console SBI functions (`sbi_console_getchar()`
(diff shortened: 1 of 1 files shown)


avpatel:
@jrtc27 and @aswaterman please check if the SBI v0.1 text updates look okay.

riscv-sbi.adoc
```diff
@@ -330,9 +330,16 @@ value for this CSR.
...
 == Legacy Extensions (EIDs #0x00 - #0x0F)
 
-The legacy SBI extensions ignores the SBI function ID field, instead being
-encoded as multiple SBI extension IDs. Each of these extension IDs must be
-probed for directly.
+The legacy SBI extensions:
+
+* Ignores the SBI function ID field, since these are encoded as multiple SBI
+  extension IDs
+* Returns 0 in `a0` register upon success
```
aswaterman, line 337: I think you need to move the return value into the definition of each function, rather than having it up here, because not all of them return 0.  console_getchar returns the unsigned character value, or -1 if no character is present.  clear_ipi returns zero if no IPI had been pending, or an implementation-defined positive value if an IPI had been pending.

avpatel: Sure, I will update like you suggested.

riscv-sbi.adoc
```diff
@@ -330,9 +330,18 @@ value for this CSR.
 
 == Legacy Extensions (EIDs #0x00 - #0x0F)
 
-The legacy SBI extensions ignores the SBI function ID field, instead being
-encoded as multiple SBI extension IDs. Each of these extension IDs must be
-probed for directly.
+The legacy SBI extensions follow a slightly different binary encoding as
```
jrtc27, line 333: This is more than just "binary encoding"

riscv-sbi.adoc
```diff
@@ -330,9 +330,18 @@ value for this CSR.
...
+
+* The SBI function ID field in `a6` register is ignored because these are
+  encoded as multiple SBI extension IDs.
+* Nothing is returned in `a1` register.
+* All registers except `a0` must be preserved across an SBI call by the
+  callee.
+* The value returned in `a0` register is SBI legacy extension specific.
+
+The page and access faults taken by the SBI implementation while accessing
+memory on behalf of the supervisor are redirected back to the supervisor.
```
jrtc27, line 344: Crucially _with sepc referring to the ecall instruction_, as opposed to the actual load/store in M-mode

jrtc27: (which makes the traps look extremely weird, and means the calls must be resumable)

avpatel: Okay, I will add details about sepc referring to the ecall instruction

atishp04 (approved):

PR #110, opened 2023-04-24
fix: ambiguous and invisible characters

jakezhu9:
This pull request fixes some characters that may be causing confusion or not showing up correctly:

- Replaces ’ with '
- Replaces “” with "
- Replaces non-breaking spaces (\u00A0) with regular spaces (\u0020)

### DIFF
--- a/riscv-sbi.adoc
+++ b/riscv-sbi.adoc
@@ -140,7 +140,7 @@ image::riscv-sbi-intro1.png[width=524,height=218]
 .RISC-V System with H-extension
 image::riscv-sbi-intro2.png[width=800,height=350]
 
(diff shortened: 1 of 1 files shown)


atishp04 (commented):
I think this is just keyboard layout thing. I am not sure if they actually change the meaning of the word/sentence.

jakezhu9:
> I think this is just keyboard layout thing. I am not sure if they actually change the meaning of the word/sentence.

You're absolutely right! It's only a small issue and won't really affect meaning of the sentence. However, I think that a well-written document should have consistency in punctuation and formatting throughout. If you feel like this is not an important matter, just go ahead and close this PR.

PR #117, opened 2023-07-24
Update the replacement for legacy console SBI functions

lbmeng:
Update the replacement for legacy console SBI functions

The legacy console SBI functions are now replaced by the new DBCN
extension. Update the out-of-date description for this.


atishp04 (approved):
Thanks.

PR #118, opened 2023-07-25
Correct 32-bit type values

lbmeng:
Correct 32-bit type values

HSM Hart Suspend Types, SRST System Reset Types and System Reset Reasons,
SUSP System Sleep Types are all 32-bit wide, so there is no such a value
that can be larger than 0xFFFFFFFF. Correct them.


atishp04 (approved):

atishp04:
Thanks!

PR #120, opened 2023-09-05
Few changes before SBI v2 public review

avpatel:
This pull request does few changes and clean-ups before SBI v2 goes to public review.


Makefile
jones-drew, line 1: How about tidying up the root directory by organizing files like the isa manual does

```
src/*.{adoc,ditaa}
src/images/*.png
```

jones-drew: Also, we may now want to provide a diff.orderFile file and instructions on how to use it with git-format-patch/diff/show

avpatel: Okay, let me add sub-directories for images and sources

riscv-sbi.adoc
```diff
@@ -1862,8 +1869,9 @@ The possible error codes returned in `sbiret.error` are shown in
 [cols="2,3", width=90%, align="center", options="header"]
 |===
 | Error code              | Description
-| SBI_SUCCESS             | firmware counter read successfully.
-| SBI_ERR_INVALID_PARAM   | The `flags` parameter is not zero.
+| SBI_SUCCESS             | Shared memory was set or cleared successfully.
```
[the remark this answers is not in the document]
atishp04: Oops. Thanks!

atishp04 (commented):
I added it for easier review while sharing. However, we have a CI which will auto generate the pdf while drafting a release. We can delete this.

riscv-sbi.adoc
```diff
@@ -1817,11 +1818,17 @@ struct sbiret sbi_pmu_snapshot_set_shmem(unsigned long shmem_phys_lo,
...
-shared memory physical address. The `shmem_phys_lo` MUST be 4096 bytes
-(i.e. page) aligned. The shared memory size must be 4096 bytes. The layout of
-the shared memory is described in <<table_snapshot_shmem_layout>>.
+Set and enable the PMU snapshot shared memory.
+
+If both `shmem_phys_lo` and `shmem_phys_hi` parameters are not all-ones
+bitwise then `shmem_phys_lo` specifies the lower XLEN bits and `shmem_phys_hi`
+specifies the upper XLEN bits of the snapshot shared memory physical base
+address. The `shmem_phys_lo` MUST be 4096 bytes (i.e. page) aligned and
+the size of the snapshot shared memory is assumed to be 4096 bytes. The layout
```
atishp04: Should we user a normative text here ?

PR #121, opened 2023-09-05
Makefile: Avoid diff error

jones-drew:
When making a fresh build the process will output something like

  diff: .../autogenerated/revision.adoc-snippet: No such file or directory

Avoid the error by checking for the file first.


avpatel:
Looks good to me.

atishp04 (approved):

PR #122, opened 2023-09-06
Various cleanups

atishp04:
There are usage of "is assumed to be" for normative statements. The meaning of it subjective and ambiguous. As the the intention was to use normative verbs anyways, lets just use "must" to provide more clarification.


avpatel:
Looks good to me.

atishp04:
I have updated the MR with few more changes. PTAL.

avpatel:
All commits look good.

PR #128, opened 2023-10-17
Fix the wording for remote fence address ranges

SiFiveHolland:
The two parameters are the start and size of the address range, not the start and end of the range.


atishp04 (approved):
Thanks for the fix.

PR #129, opened 2023-10-29
Define configuration of the mseccfg CSR

xypron:
Define that M-mode SBI implementations shall set

  mseccfg.sseed = 1
  mseccfg.useed = 0

if ISA extension Zkr is available.

This allows S- and HS-mode software to implement a random number generator.


atishp04:
@xypron : I don't think SBI specification is the right place to have these implementation specific details. Ideally, the upcoming platform spec or kernel booting guide is the correct place to document this expect from the firmware.

PR #155, opened 2024-07-21
Clarify the description of the `set_timer` function

jamestiotio:
Improve the phrasing on the clearing of the pending timer interrupt bit when timer interrupts are masked.

Additionally, clarify that this function can never fail.

CC: @atishp04 @jones-drew

### DIFF
--- a/src/ext-time.adoc
+++ b/src/ext-time.adoc
@@ -11,13 +11,18 @@ struct sbiret sbi_set_timer(uint64_t stime_value)
 ----
 
 Programs the clock for next event after *stime_value* time. *stime_value*
-is in absolute time. This function must clear the pending timer interrupt
-bit as well.
+is in absolute time.
 
 If the supervisor wishes to clear the timer interrupt without scheduling
-the next timer event, it can either request a timer interrupt infinitely
-far into the future (i.e., (uint64_t)-1), or it can instead mask the timer
-interrupt by clearing `sie.STIE` CSR bit.
+the next timer event, it may request a timer interrupt infinitely far
+into the future (i.e., (uint64_t)-1). Alternatively, to not receive timer
+interrupts, it may mask timer interrupts by clearing the `sie.STIE` CSR bit.
+
+This function must clear the pending timer interrupt bit when
+*stime_value* is set to some time in the future, regardless of whether
+timer interrupts are masked or not.
+
+This function always returns SBI_SUCCESS in `sbiret.error`.
 
 === Function Listing
 
--- a/src/changelog.adoc
+++ b/src/changelog.adoc
@@ -1,6 +1,7 @@
 == Change Log
 
 === Version 3.0-rc1
+* Clarify the description of the `set_timer` function
 * Added SBI DBTR extension
 * Added SBI FWFT extension
 * Added SBI SSE extension


aswaterman (commented):
This isn't a clarification, it's a change.

jamestiotio:
@aswaterman Hmm, in that case, can I kindly check on what is the current expected behavior? More concretely:
- If the timer interrupt is masked, is it expected that the pending timer interrupt bit is cleared?
- What are the expected `sbiret.value` and possible `sbiret.error` values when the `set_timer` function is called?

Or did you mean that explicitly specifying these in the spec is a significant change? I used the phrase "may not" for the pending interrupt bit clearing and the possible error codes seem to make sense, but do correct me if I am wrong or if I assumed something along the way.

aswaterman:
The way the current spec is written, it’s clear that supervisor-level software can rely on the pending interrupt being cleared regardless of whether it’s masked. So this is a backwards-incompatible change from supervisor-level software’s perspective.

Whether this is a “significant” change depends on factors I can’t personally speak to, but it’s still a change, and therefore it’s not a clarification.

avpatel:
I agree with @aswaterman .

Allowing supervisor software to set `stime_value` in the past is intentional because supervisor software may want to the time interrupt to trigger immediately.

Overall, NACK to this PR from my side.

jamestiotio:
> The way the current spec is written, it’s clear that supervisor-level software can rely on the pending interrupt being cleared regardless of whether it’s masked.

Ah got it, it was not clear on my side about this. It seems that OpenSBI is not implementing this though, at least when I was testing it with my `kvm-unit-tests` patches. When the timer interrupt was masked, the `sip.STIP` bit was not cleared. I would assume that this is a bug with OpenSBI instead, so I will first confirm that I wrote the test right and we can put this in a separate discussion.

> Allowing supervisor software to set `stime_value` in the past is intentional because supervisor software may want the timer interrupt to trigger immediately.

I was also not aware of this. In that case, let me remove that part from this PR.

aswaterman:
Defining this call to possibly fail is also backwards-incompatible. I think it’s best to close this PR.

jamestiotio:
Hmm does that mean that `set_timer` can never fail?

In that case, then I agree that this PR should be closed. Thank you for all of the clarifications. I will incorporate this feedback when I am writing my tests for `kvm-unit-tests`.

jones-drew:
I think some clarification is necessary. We shouldn't allow a nonzero sbiret.error to be set and still consider the SBI call to be successful, so I think explicitly stating that sbiret.error must be SBI_SUCCESS is necessary (even it if it's technically not a backward compatible change).

For the masking clarification, I think the "either...or..." wording of the second paragraph implies that masking the timer interrupt will "clear the timer interrupt". Changing the wording to something like

"""
If the supervisor wishes to clear the timer interrupt without scheduling the next timer event, it can request a timer interrupt infinitely far into the future (i.e., (uint64_t)-1).  Alternatively, timer interrupts can be masked by clearing the sie.STIE CSR bit.
"""

jamestiotio:
As recommended by @jones-drew, I have reopened this PR.

Since the intention has always been to clarify the description, I have made some adjustments to the phrasing of the sentences that involve clearing the pending timer interrupt bit.

As I am not very familiar with the language of formal specifications, any feedback is welcome and greatly appreciated. Thank you!

src/ext-time.adoc
```diff
@@ -11,13 +11,29 @@ struct sbiret sbi_set_timer(uint64_t stime_value)
...
 Programs the clock for next event after *stime_value* time. *stime_value*
-is in absolute time. This function must clear the pending timer interrupt
-bit as well.
+is in absolute time.
 
 If the supervisor wishes to clear the timer interrupt without scheduling
-the next timer event, it can either request a timer interrupt infinitely
-far into the future (i.e., (uint64_t)-1), or it can instead mask the timer
-interrupt by clearing `sie.STIE` CSR bit.
+the next timer event, it can request a timer interrupt infinitely far
```
atishp04, line 17: /s/can/may

src/ext-time.adoc
```diff
@@ -11,13 +11,29 @@ struct sbiret sbi_set_timer(uint64_t stime_value)
...
+into the future (i.e., (uint64_t)-1). Alternatively, to not receive timer
+interrupts, they can be masked by clearing `sie.STIE` CSR bit.
```
atishp04, line 19: Or it may mask the timer by clearing the `sie.STIE` CSR bit.

src/ext-time.adoc
jamestiotio: Done

src/ext-time.adoc
jamestiotio: Done

src/ext-time.adoc
```diff
@@ -11,13 +11,29 @@ struct sbiret sbi_set_timer(uint64_t stime_value)
...
+
+This function must clear the pending timer interrupt bit when
+*stime_value* is set to some time in the future, regardless of whether
+timer interrupts are masked or not.
+
+`sbiret.value` is set to zero and the possible error codes returned in
```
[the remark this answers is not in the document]
jones-drew: I also never submit this 3 day old review comment until now which referred to the 3 week old comment...

src/ext-time.adoc
```diff
@@ -17,7 +17,22 @@ bit as well.
 If the supervisor wishes to clear the timer interrupt without scheduling
 the next timer event, it can either request a timer interrupt infinitely
 far into the future (i.e., (uint64_t)-1), or it can instead mask the timer
-interrupt by clearing `sie.STIE` CSR bit.
+interrupt by clearing `sie.STIE` CSR bit. If the timer interrupt is masked,
+the pending timer interrupt bit may not be cleared.
+
+`sbiret.value` is set to zero and the possible error codes returned in
+`sbiret.error` are shown in <<table_time_set_timer_errors>> below.
```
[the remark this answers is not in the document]
jamestiotio: Done

src/ext-time.adoc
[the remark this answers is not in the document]
jamestiotio: Done

jones-drew (approved):
Reviewed-by: Andrew Jones <ajones@ventanamicro.com>

atishp04 (approved):

PR #156, opened 2024-07-24
Clarify SBI IPI and RFENCE error codes

jamestiotio:
Add SBI_ERR_INVALID_PARAM as a possible error code to IPI and RFENCE functions which take in `hart_mask` and `hart_mask_base` input parameters. While this is specified in the Binary Encoding chapter, explicitly including it in the function descriptions would make it harder to miss. SBI_ERR_INVALID_PARAM is also specified for invalid `asid` and `vmid` parameters in some RFENCE functions.

Furthermore, also add SBI_ERR_FAILED as the default error code for any other unspecified or unknown failure reasons.

CC: @atishp04 @jones-drew


atishp04:
lgtm.

jones-drew:
I agree with this. SBI_ERR_INVALID_PARAM can also be specified to apply to asids and vmids for the rfence functions, which the binary encoding section doesn't consider. Also, adding SBI_ERR_FAILED "The reset request failed for unspecified or unknown other reasons." gives SBI implementations a catchall if they have cases not considered by the spec. I recommend we add that as well.

avpatel:
Please update the changelog for 3.0-rc1 as part of this PR

jamestiotio:
I have incorporated @jones-drew's suggestions and updated the changelog as requested by @avpatel in the latest commit.

atishp04:
Resolved the conflict with change log and did a manual merge.

PR #160, opened 2024-08-08
159 fix spec builds local and gh action

wmat:
Updated build to use Docker both for local builds and the GH Action build.


avpatel:
This PR simply throws away all the Makefile improvements that we had done in the past which is not acceptable.

wmat:
This is how most of the other specs in the risc-v family of documents are setup.  However, if you'd prefer to stick with your Makefile setup we'll have to find another way to fix the build.

avpatel:
@wmat I have another PR (https://github.com/riscv-non-isa/riscv-sbi-doc/pull/162) to fix this issue which tries to keep the functionality in Makefile as much as possible.

wmat:
Cool.  Feel free to kill off my PR.

PR #164, opened 2024-09-09
Add SBI_ERR_FAILED to several functions

jones-drew:
Many functions need SBI_ERR_FAILED with the generic "...failed for
unspecified or unknown other reasons" meaning since SBI implementations
can fail for many [SBI-specific] reasons and it's not reasonable to try
and enumerate them all. A quick audit found the main need is by
functions which take physical address parameters since SBI implementations
may fail on addresses which S-mode knows to be valid (per the platform's
memory layout and ISA), but are not valid for the specific SBI
implementation. Also during the quick audit one SSE function was
identified as needing SBI_ERR_FAILED since event injection has lots of
moving parts.


atishp04:
Closing the PR as the patches are merged from the mailing list.

PR #172, opened 2024-10-16
Fix typo in Chapter 17.8 and 17.7: Read/Write software event attributes (SSE ext)

sfiedler0:
Previously "paramter", now "parameter".


pathakraul:
What about in the already ratified extensions, there are couple of typos in PMU extension i found

atishp04:
We can always fix typos in ratified spec. Please send a PR.

Issue #158, opened 2024-08-08
Value overflow on sbiret? 


leokolln:
**Binary Encoding** chapter defines `sbiret.value` as `long`.

Extension **DBCN** functions _Console Write_ and _Console Read_ first parameter `num_bytes` is `unsigned long`.

The number of bytes written/read is returned in `sbiret.value`.

If a value greater than `LONG_MAX` is used as argument for `num_bytes` , what is the expected behavior?

- Overflow on `sbiret.value`?
- Caller should clamp the argument to `LONG_MAX`?
- Callee should only operate up to `LONG_MAX` bytes per call?
- Caller should consider **Binary Encoding**'s definition of `struct sbiret{}` only as a general suggestion, but each extension can have an actual different definition of it?
- Something else?

I did not check the remaining documentation to see if other such cases exists...

atishp04:
The spec also says
"This is a non-blocking SBI call and it may do partial/no writes if the debug console is not able to accept more bytes."

Thus, the caller is supposed to verify the number of bytes returned and make multiple calls if it doesn't match num_bytes.

That's why #3 is what callee should do. 
Cc: @avpatel @jones-drew : We should add this to the kvm-unit-test DBCN test.

I checked the openSBI implementation which allows more than LONG_MAX. So that should be fixed if my understanding above is correct.

jones-drew:
> The spec also says "This is a non-blocking SBI call and it may do partial/no writes if the debug console is not able to accept more bytes."
> 
> Thus, the caller is supposed to verify the number of bytes returned and make multiple calls if it doesn't match num_bytes.
> 
> That's why #3 is what callee should do. Cc: @avpatel @jones-drew : We should add this to the kvm-unit-test DBCN test.
> 

Yup, I got Cade to implement the basic write buffer test with a loop on num_bytes. Also, I started writing a couple more tests to see what happens when we cross page boundaries and, for 32-bit, what happens when we use addresses > 4G or cross from under 4G address ranges into 4G address ranges (that found a couple opensbi bugs that I still need to patch)

> I checked the openSBI implementation which allows more than LONG_MAX. So that should be fixed if my understanding above is correct.

The spec doesn't say anything about limits on num_bytes, which means up to ULONG_MAX being allowed is implied. So SBI implementations should ensure that base_addr + num_bytes isn't greater than HIGHEST_PHYSICAL_ADDRESS and return INVALID_PARAM if it is.

[...]

PR #184, opened 2024-11-08
Clarify the sbiret.value

atishp04:
As per the calling convention a1 will have sbiret.value on ecall return. However, the current specification describes an anlogous C structure describing as long. Some of the SBI extension may require an "unsigned long" data type instead of "long".

Relax the analgous C structure definition to allow that.

Fixes: https://github.com/riscv-non-isa/riscv-sbi-doc/issues/158


atishp04:
Cc: @avpatel @jones-drew @SiFiveHolland

avpatel:
Looks good to me.

PR #185, opened 2024-11-08
Fix the table reference in set/get functions

atishp04:


avpatel:
Looks good to me.

jones-drew:
Ack, thanks!

PR #189, opened 2024-12-10
DBTR typos

pdonahue-ventana:
Typo fixes.

Also, clarify that all harts don't have one trig_max. Each has a separate trig_max.


SiFiveHolland:
LGTM.

PR #195, opened 2025-01-21
Address SBI mpxy ARC feedback

avpatel:


src/ext-mpxy.adoc
```diff
@@ -655,19 +653,24 @@ sbi_mpxy_send_message_without_response(uint32_t channel_id,
...
+message protocol specification.
 
-In cases where complete data transfer requires multiple transmissions, the
-supervisor software shall issue multiple requests as necessary. Details of such
-cases can be found in the respective message protocol specification.
+NOTE: The messages which do not have an expected response as per the underlying
+message protocol specification are also referred to as posted messages. This
+function should be only used for such posted messages and the respective message
+protocol specification should define a mechanism to track the status of posted
+messages using notification events or some other message with response.
```
atishp04, line 689: The 2nd part is optional - Correct ? Can we clarify that ? Something like this:
This function should be only used for such posted messages. The respective message
protocol specification should define a mechanism to track the status of posted
messages using notification events or some other message with response if required.

src/ext-mpxy.adoc
```diff
@@ -275,25 +276,43 @@ assigned ID.
...
-=== Function: Set shared memory (FID #0)
+=== Function: Get shared memory size (FID #0)
 
 ```c
-struct sbiret sbi_mpxy_set_shmem(unsigned long shmem_size,
-                                 unsigned long shmem_phys_lo,
+struct sbiret sbi_mpxy_get_shmem_size(void)
+```
+
+Get the shared memory size for sending and receiving messages. The shared memory
```
atishp04, line 285: shared memory size in number of bytes ?

atishp04:
Apart from these nit comments, looks good to me.

src/ext-mpxy.adoc
avpatel: Okay, I will update.

src/ext-mpxy.adoc
avpatel: Okay, I will update.

avpatel:
@atishp04 I have addressed your comments. The build failure is due to some other issue.

atishp04 (approved):

PR #223, opened 2025-09-25
ext-sse: add local unknown nmi event id

cuiyunhui:
Take ID value 0xffff8001 from the "Global events reserved for future use" range as the SBI_SSE_EVENT_GLOBAL_UNKNOWN_NMI event ID.


Ziang95 (approved):
LGTM

AemonZhang:
I'm curious about the reasoning behind choosing this specific ID value. Since all events have their priority set to 0 by default, the UNKNOWN_NMI event would end up with a lower priority than other known events (e.g., PMU). Why didn't we pick an ID from a higher-priority range instead, such as 0x00000002?

PR #259, opened 2026-03-19
Antora refactor

wmat:
Merges in the Antora requirements so that the spec can be rendered for docs.riscv.org


MaskRay:
This PR was merged without any review or approval from the repository maintainers, which is not the appropriate process for changes to a shared spec repository. Could a repo admin please force-push `master` to remove these commits (i.e. reset to the pre-merge state at the commit before `2e197d1`) to prevent users to get large binary artifacts like `modules/ROOT/attachments/riscv-sbi.pdf`?

avpatel:
@aswaterman @wmat I have dropped https://github.com/riscv-non-isa/riscv-sbi-doc/commit/9d66b1585256e4ec4b31bdb27aae00902222bb1b and force pushed 

Also, SBI spec builds fine after the force push.

wmat:
Thanks @awaterman and sorry again for the noise.

PR #261, opened 2026-04-25
doc: clarify 4-byte alignment requirement for start_addr

guoren83:
Linux can pass a non-4-byte-aligned address to sbi_hsm_start() [1]. To prevent this invalid usage, we now explicitly document that `start_addr` must be 4-byte-aligned (matching the stvec/vstvec BASE requirement).

[1]: https://lore.kernel.org/linux-riscv/20260413132009.133752-1-cp0613@linux.alibaba.com/

### DIFF
--- a/src/ext-hsm.adoc
+++ b/src/ext-hsm.adoc
@@ -137,6 +137,8 @@ The possible error codes returned in `sbiret.error` are shown in the
                               * Executable access to the address is prohibited
                                 by a physical memory protection mechanism or
                                 H-extension G-stage for supervisor-mode.
+                              * It is not 4-byte-aligned, as required by the
+                                stvec BASE address alignment requirement.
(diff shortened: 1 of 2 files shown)


radimkrcmar:
This change is backward-incompatible.  The Linux patch you posted is a clear example that legal software might depend on a 2-byte aligned address.

There also doesn't really seem to be a reason for the constraint.   The entry method is going to be mret/sret, so stvec/vstvec shouldn't really be coming into the picture.  The stvec/vstvec registers are in an unspecified state upon entry, and a value of 0 seems more appropriate.

We should fix OpenSBI to conform to the specification instead.

avpatel:
There is no relation of `start_addr` parameter of `sbi_hsm_start()` with stvec CSR. In fact, SBI implementation (M-mode or HS-mode) use mepc or sepc CSRs to start a hart at particular `start_addr`.

PR #262, opened 2026-06-20
Add SBI watchdog extension

avpatel:
Define a new SBI watchdog (WDT) extension which allows SBI implementation to provide watchdog functionality for the supervisor software.


src/ext-watchdog.adoc
```diff
@@ -0,0 +1,334 @@
...
+the supervisor software using `sbi_watchdog_pat()` before the current
+watch period expires, indicating normal operation.
+
+Upon system events (such as system suspend or external debug) where
+the supervisor software will not get chance to re-start or pat the
+watchdog timer, the SBI implementation should transition the watchdog
+timer from `RUNNING` to `PAUSED` state and pause counting. After
+resuming normal operation, the SBI implementation should transition
+back the watchdog timer from `PAUSED` to `RUNNING` state and resume
+counting.
```
radimkrcmar: Don't we want software knowledge and ideally also control over the pause behavior in suspend -- for example to use the watchdog as a wake-up source?

(Also, I think the second `should` would be a bit clearer as `must` if SBI pauses the watchdog, because leaving the watchdog in PAUSED state after suspend/debug would make it unusable to software.)

src/ext-watchdog.adoc
```diff
@@ -0,0 +1,334 @@
...
+The supervisor software can optionally subscribe for a watchdog
+notification which is delivered as a message signaled interrupt
+(MSI) before the current watch period expires while the watchdog
+timer is in the `RUNNING` state. The supervisor software can also
+optionally subscribe for a watchdog expiry event which is delivered
+as a supervisor software event (SSE) after the watchdog timer moves
+to the `EXPIRED` state and before initiating the system reset. After
+delivering watchdog expiry event, the SBI implementation must wait
+1 second before initiating the system reset to allow the supervisor
+software handle the watchdog expiry event. The <<table_sse_event_ids>>
```
radimkrcmar: The SSE handler can't really do anything with the watchdog anymore and system will reset in 1 second.
I think it might be useful to have the SSE handler save the system -- perhaps perform a kexec and dump, which might take more than a second.

We could add a SSE timeout attribute, so the SRST is immediate on EXPIRED, but software can register the SSE a second earlier.  It allows the current behavior, and also gives flexibility to achieve more.

src/ext-watchdog.adoc
```diff
@@ -0,0 +1,334 @@
...
+
+| WDT_MAX_PERIOD
+| 0x00000003
+| RO
+| Maximum watch period duration (in seconds).
+
+| WDT_PERIOD
+| 0x00000004
+| RW
+| Watch period duration (in seconds). +
```
radimkrcmar: I think that seconds are too coarse.  Real-time systems might want to reset in less than one second.

Is there a reason to not use at least milliseconds?

src/ext-watchdog.adoc
avpatel: Watchdog is almost never used as wake-up source. We have introduced a separate `PAUSED` state to ensure that start(), stop(), and pat() don't work when watchdog is paused.

Regarding the second `should`, I agree it to be `must`. I will update.

src/ext-watchdog.adoc
SiFiveHolland: What happens if the SSE handler calls `sbi_watchdog_pat()` or `sbi_watchdog_stop()`?

If the SSE handler returns more quickly than 1 second, should the SRST happen immediately?

src/ext-watchdog.adoc
SiFiveHolland: Yes, or even microseconds, which still allows a timeout longer than an hour. A higher precision means more flexibility for the users of the API.

avpatel: The Linux Watchdog interface expects all durations in seconds hence the SBI watchdog extension using second granularity.

src/ext-watchdog.adoc
avpatel: The purpose of SSE handler is to take a dump of whatever it can for debug purpose but we can't leave the timeout to be open-ended. Either the SSE timeout is implementation specific or we need an attribute which is only writeable in `STOPPED` state.

src/ext-watchdog.adoc
```diff
@@ -0,0 +1,334 @@
...
+
+| WDT_MIN_PERIOD
+| 0x00000002
+| RO
+| Minimum watch period duration (in seconds).
+
+| WDT_MAX_PERIOD
+| 0x00000003
+| RO
+| Maximum watch period duration (in seconds).
```
SiFiveHolland: Since this is a software construct, why would the SBI watchdog not support any period up to UINT_MAX?

src/ext-watchdog.adoc
```diff
@@ -0,0 +1,334 @@
...
+
+| WDT_NOTIF_TIME
+| 0x00000006
+| RW
+| Time (in seconds) from the start of watch period after which
+  watchdog notification is received by the supervisor software. +
+  +
+  The value of this attribute must be less than the current watch
+  period duration `WDT_PERIOD` and a value `0` of this attribute
+  implies watchdog notifications are disabled. +
```
SiFiveHolland, line 135: What is the behavior when the watchdog period is changed? Would it make more sense to define this as a time before expiration?

src/ext-watchdog.adoc
```diff
@@ -0,0 +1,334 @@
...
+                          `attribute_id` parameter is reserved.
+| SBI_ERR_FAILED        | Failed due to other unspecified errors.
+|===
+
+=== Function: Write watchdog attribute (FID #1)
+
+[source, C]
+----
+struct sbiret sbi_watchdog_write_attribute(uint32_t attribute_id,
+                                           uint32_t attribute_value)
```
SiFiveHolland, line 241: Since the attribute-based API only allows reading or writing one attribute at a time, why not make these separate SBI functions, and save the level of indirection? Additionally, having a separate function would allow setting all MSI parameters at once.

src/ext-watchdog.adoc
radimkrcmar: > What happens if the SSE handler calls sbi_watchdog_pat() or sbi_watchdog_stop()?

The ecalls fail with `SBI_ERR_INVALID_STATE`, because the watchdog wasn't in `RUNNING` state.

> If the SSE handler returns more quickly than 1 second, should the SRST happen immediately?

The SSE handler could SRST itself if it wanted to, so specifying that seems unnecessary.

I think the only reason why the SSE handler would `sbi_sse_complete` is to continue execution after fixing the hang (e.g. configuring an environment for a successor kernel) and the watchdog shouldn't SRST then.

> The purpose of SSE handler is to take a dump of whatever it can for debug purpose but we can't leave the timeout to be open-ended.

The `can't` assumes that the SSE handler is corrupted and keeps the system alive by mistake, which should be about as probable as that the pat handler is corrupted and keeps a hung system alive, so I don't think that requiring reset after the SSE notification is improving the safety guarantees.

> Either the SSE timeout is implementation specific or we need an attribute which is only writeable in STOPPED state.

A user-controllable SSE timeout after WDT_PERIOD expires should be mostly sufficient if it can be extremely long.

src/ext-watchdog.adoc
radimkrcmar: Linux watchdog framework internally handles ms for the attributes, so the extra precision wouldn't be lost.  For example when userspace requests longer timeout than the watchdog can handle.
```
        unsigned int min_hw_heartbeat_ms;
        unsigned int max_hw_heartbeat_ms;
```

Multiplication because of higher precision is cheap.  We could consider specifying that the watchdog is allowed to round the values requested by S-mode to the granularity of watchdog ticks, to avoid excluding awkward implementations.

src/ext-watchdog.adoc
avpatel: > The `can't` assumes that the SSE handler is corrupted and keeps the system alive by mistake, which should be about as probable as that the pat handler is corrupted and keeps a hung system alive, so I don't think that requiring reset after the SSE notification is improving the safety guarantees.

I think you misunderstood the purpose of SSE handler. The SSE handler is a debugging feature and only allows supervisor software take dump of registers. By the time SSE handler is called, the watchdog timer is already in `EXPIRED` state and the `sbi_watchdog_pat()` won't work from the SSE handler so SSE handler can't keep the system alive.

In other words, the system will be rebooted irrespective SSE handler returns or not.

> 
> > Either the SSE timeout is implementation specific or we need an attribute which is only writeable in STOPPED state.
> 
> A user-controllable SSE timeout after WDT_PERIOD expires should be mostly sufficient if it can be extremely long.

I am thinking of adding WDT_SSE_TIMEOUT attribute which is only writeable when the watchdog timer is in `STOPPED` state. The default value of this attribute will be 1 second.

src/ext-watchdog.adoc
avpatel: I think microseconds granularity is fine because it indirectly limits maximum watch period to roughly 1.19 hour.

src/ext-watchdog.adoc
avpatel: I agree, let's drop `WDT_MAX_PERIOD`

src/ext-watchdog.adoc
avpatel: If the `WDT_PERIOD` changes while watchdog is in `RUNNING` then new value of `WDT_PERIOD` will be used from the next watch period.

I choose watchdog notification time from the start of watch period since that seemed more natural to be but I don't mind changing this to be from the end of watch period.

src/ext-watchdog.adoc
avpatel: A separate read/write SBI call for each attribute would mean a lot of SBI functions where most of these attributes are only configured at driver probe time.

I do agree that supervisor software should have mechanism to program all MSI parameters in SBI call so let me improve the `sbi_watchdog_read/write_attribute()` to allow accessing multiple attributes in one SBI call.

src/ext-watchdog.adoc
```diff
@@ -0,0 +1,350 @@
...
+| Minimum watch period duration (in microseconds).
+
+| WDT_PERIOD
+| 0x00000003
+| RW
+| Watch period duration (in microseconds). +
+  +
+  The value of this attribute must be greater than or equals minimum
+  watch period duration and it must be less than or equals maximum
+  watch period duration. +
```
SiFiveHolland, line 110: This description still references the maximum duration attribute, which no longer exists.

By the same logic as my previous comment (I should have been clearer at the time), is there any situation where `WDT_MIN_PERIOD` would need to be any value other than 1? We don't guarantee that the reset completes within a specific bounded time from entering the `EXPIRED` state, so even if minimum hardware timer/hardware watchdog resolution is larger than the requested `WDT_PERIOD`, the implementation would still be compliant. (This is important if the SBI implementation is, say, a hypervisor in a scheduled-out supervisor domain.) That said, I can see why it might be surprising if the watchdog expires significantly later than the requested period.

avpatel: > This description still references the maximum duration attribute, which no longer exists.

Okay I will update.

> 
> By the same logic as my previous comment (I should have been clearer at the time), is there any situation where `WDT_MIN_PERIOD` would need to be any value other than 1? We don't guarantee that the reset completes within a specific bounded time from entering the `EXPIRED` state, so even if minimum hardware timer/hardware watchdog resolution is larger than the requested `WDT_PERIOD`, the implementation would still be compliant. (This is important if the SBI implementation is, say, a hypervisor in a scheduled-out supervisor domain.) That said, I can see why it might be surprising if the watchdog expires significantly later than the requested period.

The SBI watchdog is typically implemented on-top-of some physical timer who's tick granularity is the minimum watch period so `WDT_MIN_PERIOD` can't be removed since it represents underlying hardware limitation. For example, if a platform has mtimer running at 1 MHz then `WDT_MIN_PERIOD` is 1 microseconds whereas if a platform has mtimer running at 1 KHz then `WDT_MIN_PERIOD` is 1000 microseconds. The SBI implementation may also intentionally want minimum watch period to be at least above certain value to avoid too frequent watchdog patting.

